Issue #2, opened 2017-03-13
Versioned guides


samselikoff:

samselikoff:
@dfreeman wrote

> Can you elaborate on how you imagine the versioned guides working? Would each version be a separate dummy app, or do you imagine them coexisting within a single application? Separate apps seem like they may be thorny to maintain and deploy, but if the guides include live (and possibly editable) examples, having multiple versions coexist in a single application could be difficult.

I haven't thought much about this yet.

Mirage has versioned docs, and I do this by literally [copy-pasting the guides directory](https://github.com/samselikoff/ember-cli-mirage/tree/gh-pages/docs) for each version. It's a Jekyll app, but you could imagine doing something similar in the Ember dummy app.

Now on the flip side, we will be needing to solve the versioned API docs problem as well. And the source for the API docs will certainly live in different versions of the project - since they'll be pulled from the actual tagged releases of the source code.

Seems we can either make everything static, or put things on a server somewhere.

I started to write more but I need more time to think about this. Back later. If anyone has thoughts here drop a line.

dfreeman:
Sounds good. I'll let this rattle around in the back of my head while I work on other stuff and post back if anything occurs to me.

samselikoff:
For example I will use this to rewrite Mirage's docs site. Want to port over the old versions and probably put them in a static place, public? Have to think about this.

dfreeman:
So I just took some time to page [the long conversation we had back in April](https://embercommunity.slackarchive.io/-team-learning/page-21/ts-1492134424735458) back into my brain. It spans a bunch of topics, but I think has some relevance here. My biggest takeaway was that in order to support interactive guides across different versions of the addon in question, our only real option was going to be multiple deployed Ember apps.

If we were to go with that approach treat `gh-pages` as a essentially a place for static build output, coexistence with older versions of documentation becomes pretty straightforward — going to `my-addon/v0.1.x` could load a precompiled Jekyll page (or maybe even not prebuilt?), and `my-addon/v0.2.x` could load a deployed Ember app.

Just spitballing on a directory structure, I could imagine something like:

- `index.html` (meta redirect to latest version?)
- `docs/`
  - `v0.1.x/`
    - `index.md`
    - ...
  - `v0.2.x/`
    - `index.html`
    - `assets/`
      - ...
    - ...
- `data/`
  - `api-docs/`
    - `projects/`
      - `my-addon.json`
    - `project-versions/`
[...]

PR #4, opened 2017-03-21
First pass at generating doc JSON

dfreeman:
As a simple first cut, this produces JSON that `ember-api-docs` can more or less read. By running `ember generate-docs` in this repo, and then copying `public/docs` from here into `public/json-docs/ember-cli-addon-docs` in the viewer, I was able to load:

image

Still plenty of work to be done in terms of making the stuff in `ember-api-docs` apply more generally, capturing additional addon doc versions, and so on.

Some stream-of-consciousness notes I took as I was working on this and things occurred to me:
 - `ember-jsonapi-docs` had some hard-coded values and paths in some modules that kept me from using them here
 - `ember-api-docs` makes some assumptions about doc structure, e.g. at least one "namespace" (== `@static @class`) that may not necessarily hold true for addons
 - Here I'm generating docs with a CLI command into a `gitignore`d directory, but...
   - longer term maybe it makes sense to tie into the release process?
   - in @locks's `rubydoc.info` vision for the future, it's probably actually beneficial to check the JSON into source control
 - A shared addon repo may eventually be useful to have for developing this against, particularly when we start to look at versioned docs/guides and wouldn't want to start introducing random tags here


dfreeman:
P.S. @samselikoff I'm totally uninitiated in the ways of Github Projects, so if I could be better tracking this over in that tab somehow, please let me know 🙂

samselikoff:
Nice work! First question, so the command generates both json and the site you screenshotted above?

dfreeman:
The command just generates json; the screenshot above is the result of copying that json into [ember-api-docs](https://github.com/ember-learn/ember-api-docs)'s `public` directory, just as a proof that it's generating the structure the existing components understand.

samselikoff:
Ahh perfect. Next question, how long does the build take? Is it feasible to make it a broccoli plugin that would be part of the project build (rather than a separate command)?

Related, storing the json in the project means future rubydoc.info-clone would use github as a backend... is that feasible? Is it what rubydoc.info does?

sivakumar-kailasam:
My 2¢, if we could generate the docs generated by this app as static files that we could place inside a folder, then it would help in both the rubydoc clone as well as an independent github docs folder(for the project site).

samselikoff:
I am generally very against storing build artifacts with repos meant for source code, can lead to all sorts of unnecessary issues. We don't need it stored as static files for the github docs project, since that's going to be the dummy app build by `ember build`.

I need to understand the architecture of the rubydoc thing more.

dfreeman:
> Ahh perfect. Next question, how long does the build take? Is it feasible to make it a broccoli plugin that would be part of the project build (rather than a separate command)?

On my machine, that command spends 250-350ms generating and transforming the doc JSON for just the couple of files in `addon/`. It's feasible to do this within broccoli (and may not be totally horrendous with appropriate caching), but my guess is that it'll become somewhat unpalatable as the size/complexity of the addon grows.

> Related, storing the json in the project means future rubydoc.info-clone would use github as a backend... is that feasible? Is it what rubydoc.info does?

I don't know what rubydoc.info does, but I'm imagining this could work somewhat similarly to how EmberObserver's code search [works today](http://codeallday.com/blog/code-search-in-ember-observer/) — it basically scrapes new/updated addons daily from Github and indexes them into a Google codesearch index.

dfreeman:
One additional point against doing the doc generation from within broccoli is that it complicates the question of producing versioned docs...

samselikoff:
Interesting. Ugh... want to sleep on this, would really like to save addon maintainers from headaches associated with build artifacts in repos. If we can think of any other way...

The version docs thing may be in the nail in the coffin. Will sleep on it.

samselikoff:
Store a different Ember app under `v.0.2.x`, `v0.3.x` subfolders in `gh-pages` branch...

dfreeman:
I just had exactly that same thought 🙂

samselikoff:
And if we do rubydoc.info clone, we could provide script that a job could use to build docs from raw source. If that was running nightly and backing up somewhere it wouldn't matter.

samselikoff:
Cool thing here is if someone finds typo in `v0.1.x` docs page, they could make an edit on `0-1-stable` and redeploy docs by cping to `gh-pages/v0-1-x` without having to `git checkout` through the rest of the tagged versions.

dfreeman:
I'll sleep on it for now and continue to mull it over tomorrow, but it feels like the gh-pages branch may be the key to where we can stash this stuff. The doc JSON will have to live there anyway for the dummy app, so it's entirely possible we could come up with reasonable conventions for making that data findable by rubydoc.info-clone (needs a catchier name) as well.

samselikoff:
Oh perfect, of course - gh-pages is basically our build directory. But this is nice bc it keeps source free from build artifacts, so folks won't publish versions to npm then have to republish bc they forgot to build something.

locks:
The rubydoc.info-style documentation hub would download the gem and build the API documentation itself. If you install a rubygem you can see `ri` compiling the docs at install-time (unless you've deactivated that), for example.

locks:
I'll try giving https://github.com/docmeta/rubydoc.info a look as well!

dfreeman:
> The rubydoc.info-style documentation hub would download the gem and build the API documentation itself, I think.

That approach obviously works, but it feels like having this intermediate JSON format gives us a tool at our disposal that rubydoc.info doesn't have, and we may want to consider whether it's worth it to use that instead.

Pros for parsing from the source:
 - not dependent on the developer to actually publish their docs
 - more flexibility if we decide to change the intermediate JSON format

Pros for reading the built JSON from `gh-pages`:
 - minimal processing work required when pulling docs
 - provides a common doc "compilation target" independent of the details of the in-source documentation format

The appealing thing to me about using the JSON artifacts is the idea that would make it easier to support things like TypeDoc in the future, since the only part of the pipeline that would have to vary would be on the machine building and deploying the addon.

Then the rubydoc.info-style doc hub could be completely agnostic, since it would never need to look at the source itself.

samselikoff:
I think we should set aside the implementation details of the rubydoc-clone for now, and just make sure our architecture decisions don't cause issues down the line.

Seems like we're in agreement: from src, there will be a command that will build YUIdocs JSON (for a given version) and then copy that output into gh-pages branch, under some directory. We need to discuss the conventions for the dir structure, but I think this is the correct approach.

Any other feedback, before moving onto filesystem layout conventions discush?

locks:
Let's forge ahead! Out of hand I think we need an index `versions.json` with a list of all the versions that we should list, which also allows us to change the structure later on if we want to. The biggest restriction I can think of is we need to take versioning into account, also likely bake `/current` as a version in.

samselikoff:
`versions.json` is an awesome idea, it could even have a `current` property that points to a directory so we don't need a separate `current` directory.

`versions.json` would also allow us to change the location of the json files if we ever needed to add that. For now we could put it under something like `/data`.

dfreeman:
Big 👍  on a centralized `versions.json` for tracking what's available + current.

One question I have is whether we're _just_ deploying different versions of the data to `gh-pages`, or actually building one copy of the dummy app per version—which way we go presumably impacts the directory structure we're looking at.

The upside to having multiple copies of the app is it provides a really simple answer to #2, and it would make it possible to make the doc generation part of the broccoli flow instead of an independent command.

locks:
We can do the Guides thing of just generating the current version's docs. As you release new versions, you'll add them to the folder. I think there should be one app that can load N versions, like ember-api-docs

dfreeman:
That leaves us without a clear answer to the guides versioning, but I guess that's not the point of this thread 🙂

locks:
@dfreeman baby steps ;)

samselikoff:
Interesting... was thinking just the json data but, you're correct that a separate app per version takes care of versioned guides. Definitely some tradeoffs here, will think more about it.

dfreeman:
Based on our Slack conversation last night, I just pushed an update to generate doc JSON in `public/docs` via Broccoli (note that this is functional, but could be somewhere between a little and a lot smarter about caching).

Currently this outputs the same structure that `ember-api-docs` uses for its JSON, but there's nothing forcing that on us if we decide there's a different form factor that's more useful. See the [application adapter](https://github.com/ember-learn/ember-api-docs/blob/a1c5f33d9b612437acfec4258e2edac8263df602/app/adapters/application.js) there for a rough idea of how they're consuming these.

Where ember-cli-addon-docs is its own documentation target, the output looks something like:

```
public/docs
├── 0.0.0+4c1c3127
│   ├── classes
│   │   └── ember-cli-addon-docs-0.0.0+4c1c3127-SampleComponent.json
│   ├── namespaces
│   │   └── ember-cli-addon-docs-0.0.0+4c1c3127-MyNamespace.json
│   └── project-versions
│       └── ember-cli-addon-docs-0.0.0+4c1c3127.json
└── projects
    └── ember-cli-addon-docs.json
```

samselikoff:
@dfreeman do you think we can ship this & iterate? How do you feel about it

locks:
Yeah, I'm all for getting some momentum going :)

dfreeman:
@samselikoff Sounds good to me. After this upcoming weekend I'll finally be done traveling for quite a while, so I should be able to re-engage more then :)

PR #23, opened 2017-08-09
Move snippet and clipboard to dependencies

ming-codes:
This fixes a lot of issues people are seeing.


mikemunsie:
This will fail until ember-code-snippet supports addons. I submitted a PR on their end that will enable it. https://github.com/ef4/ember-code-snippet/pull/38 

I have tested locally and have confirmed this will work! I do have a fork of this, but let's see how if the PR gets any attention.

package.json
```diff
@@ -23,6 +23,8 @@
     "ember-ace": "^1.1.0",
     "ember-cli-autoprefixer": "^0.7.0",
     "ember-cli-babel": "^6.3.0",
+    "ember-cli-clipboard": "^0.8.0",
+    "ember-code-snippet": "^1.9.0",
```
mikemunsie: Let's move `ember-code-snippet` to 1.10.0 - the PR just got merged to support being used as an addon.

ming-codes: Done

dfreeman:
It's a little unusual that ember-code-snippet acts on behalf of the root app even when it's only included as a transitive dependency, but it works well for our use case, so I'm not complaining 🙂 

Thanks @ming-codes and @mikemunsie!

PR #27, opened 2017-08-11
Primitives for versioned deploys

dfreeman:
This change adds some primitives for managing versioned deploys. I'm happy to bikeshed on any details of the form factor, or just get it merged and iterate later.

There are two main changes here:
 - we now have a default blueprint that adds `ember-cli-deploy` and a minimal set of plugins, then produces a default `config/addon-docs.js` file
 - this addon now acts as a deploy plugin, which reads the given configuration to determine whether and where to deploy

The default configuration behaves basically the same as today, but three aspects of a potential deploy are [now programmatically configurable](https://github.com/ember-learn/ember-cli-addon-docs/pull/27/files#diff-d1cacd9ed1ac81b57e562d88fe762527), based on things like the current git commit and process enviornment:
 - whether to deploy at all
 - what directory to deploy into
 - what files in that directory (if any) to keep across deploys


samselikoff:
Been busy with a new client, but will carve out time for this soon...

samselikoff:
This looks awesome, I want to give it a try soon on a dummy repo and then we can merge it & iterate API if necessary.

Thanks for all this work!!

samselikoff:
Finally got a chance to play around with this. Awesome stuff! Here's some feedback stream-of-conciousness style:

On a fresh addon called `ember-cli-addon-test`, after installing this branch I got this

image

To fix, I did

- `git co -b gh-pages`
- `git push`

Then deploying via `ember deploy production` worked. But then I had to add some code to 
`tests/dummy/config/environment.js` under the `production` environment to get the assets to work:

```js
ENV.rootURL = '/ember-cli-addon-test';
ENV.locationType = 'hash';
```

Then I looked at the config values in the file.

- "Everything in that directory will be cleared before the new version is committed." I didn't understand this, is it saying that everything everything but the paths? Also let's add some examples, and a bit more motivating text (i.e. why would I use this)?

- "Return a boolean indicating whether or not the current deploy should actually run." Why would I use this? Examples + motivating text.

Then I checked out `gh-pages` locally to try to add a file to a `preservedPath` but my local branch was different, it wasn't tracking origin. I had to add a remote and then pull, and then `git clean -df` because I had `dist`, `node_modules` and `tmp` local and dirty. But I did this and then blew away node_modules too and then got a confusing error message (Path must be a string). Is it possible to ignore node_modules, tmp etc for our users here? Perhaps create a `.gitignore` file? (I ended up adding these dirs to my global gitignore but other users will run into this).

samselikoff:
This is all awesome and I believe will pave the way for higher level abstractions in the future, feel free to merge and we can address anything from above in other commits.

Great work again!! I really want to move mirage over to this soon.

dfreeman:
> To fix, I did
> - `git co -b gh-pages`
> - `git push`

I'd gone through the steps to set up the `gh-pages` branch and configure the dummy app so many times that I wasn't even thinking about those when I set this up. 😅

I think we can probably prompt the user as part of the installation blueprint and then automate the process of setting up the `gh-pages` branch for them. We can also add a `.gitignore` to the branch (and as a default `preservedPath`) to avoid the issue you ran into with `node_modules`, etc.

> ```js
> ENV.rootURL = '/ember-cli-addon-test';`
> ENV.locationType = 'hash';
> ```

I don't think addons have the ability to override config set in the base app; they can only provide defaults. I could imagine getting fancy and writing a little codemod for `config/environment.js`, but probably the answer for now is just documentation. Maybe emit a note when the installation blueprint runs?

> Then I looked at the config values in the file.

Yep, I definitely wrote those descriptions while my brain was still swimming with implementation details, and they're not super clear.

I'll take a pass at the comments in the generated config file to clarify them and add some motivating examples, then merge this and file a couple issues to capture the outstanding work around setting up `gh-pages` and documenting the routing configuration. Thanks for taking the time to go through this all so thoroughly!

samselikoff:
Yep, all sounds like great next steps. I can help with docs too (I think once I get the guides tidied up it will help with a few of the points I got stuck on.)

Exciting!

dfreeman:
Finally got some time to come back and finish this up. A few changes of note:
 - the generated `config/addon-docs.js` now includes some motivating examples for using each hook and reworked documentation (though I'm sure there are still improvements that could be made)
 - the `get preservePaths()` hook became `preservedPaths(info)` to mirror the form factor the other hooks
 - the default blueprint now updates your dummy `config/environment.js` to use a default rootURL of `/<addon-name>/` and a default locationType of `hash`. We might want to make the same change in the dummy app here (to make the development app more similar to the deployed one)
 - we now check at the start of a deploy if the target branch exists, and offer to create it for the user if it doesn't

Issue #9, opened 2017-07-01
Add "Edit this page" link to the top of every page


samselikoff:

samselikoff:
I am working on this feature and not sure how best to get the source filename from the current route.

For example, let's say the user is on this page:

https://ember-learn.github.io/ember-cli-addon-docs/docs/quickstart

I want the "Edit this page" link to go here:

https://github.com/ember-learn/ember-cli-addon-docs/blob/master/tests/dummy/app/pods/docs/quickstart/template.md

I poked around `requirejs.entries` and thought I might be on to something,

![image](https://user-images.githubusercontent.com/2922250/34654228-eb2bcfe8-f3c6-11e7-864b-8d1540d7c701.png)

but as you can see, the `moduleName` path is actually the `.hbs` file, which is a compiled file derived from the `.md` source file. So, that path is not correct.

Any ideas on the best way to go about doing this? Will we need to create a build-time hash that contains all module names under `docs`, then use some finding/regex to find the matching module on the client?

Also, the `docsRoutes` service already has some logic for getting the currentRouteUrl (for example `/docs/quickstart` in this case).

samselikoff:
https://github.com/ember-learn/ember-cli-addon-docs/pull/42 closes

PR #42, opened 2018-01-07
[#9] Adds Edit this page link to docs viewer

samselikoff:


index.js
```diff
@@ -186,3 +194,13 @@ function findImporter(addon) {
...
+
+class FindDummyAppFiles extends Plugin {
+  build() {
+    let addonPath = this.inputPaths[0];
+    let paths = walkSync(addonPath, { directories: false })
+    let pathsString = JSON.stringify(paths);
+
+    fs.writeFileSync(path.join(this.outputPath, 'app-files.js'), `export default ${pathsString};`);
+  }
+}
```
samselikoff: @dfreeman could use your eyes on this change. I needed a list of all the source files under `/dummy/app` to map back the final rendered templates to the original source files (since they can be either hbs or md). The plugin walks the `/tests/dummy/app` directory and creates a build-time file called `app-files.js` which is just an array of all the filenames with absolute paths. You can see how it's used by +F for "app-files".

Questions:
1. I'd have loved to be able to do this transform "inline" or using just a function, but I couldn't figure out how to do it without extending Plugin. Is there a way? My broccoli-fu is weak.
2. This was just the first thought I had for how to share this sort of static data between node and the Ember app. Is this a broccoli best practice? For this app it might not matter that much but I plan on doing more of this type of work in some other projects (transforming static data from files and making that available to Ember components at run-time). Is this the best way to share this kind of data?

dfreeman: 1. The closest I can come to doing this inline would just be `dummyAppFiles = new class extends Plugin { ...`. For cases like transforming individual files, `broccoli-stew` has helper methods like `map(tree, transformFn)`, but I don't know of one that takes a function to just operate over a whole tree.

2. The two approaches I see for sharing data like this are either a) exactly what you've done here, or b) stashing it in `config` and letting the CLI package it up for you. For a), I see [`broccoli-flatiron`](https://github.com/buschtoens/broccoli-flatiron) get a lot of use, but since you don't care about the actual file contents in this case, what you have seems pretty reasonable.

samselikoff: sweet! Thanks dude

dfreeman:
🎉

PR #43, opened 2018-01-26
Magic attributes for markdown code snippets?

dfreeman:
@samselikoff I wanted to get your take on this idea — I added support for a `title` property on `{{docs-snippet}}` to help eliminate the need for all the leading comments describing what file the snippet is tied to, so for instance this:

```hbs
{{#code-snippet name='foo'}}
  {{! app/templates/components/my-component.hbs }}
  ...
{{/code-snippet}}
```

Can become this:

```hbs
{{#code-snippet name='foo' title='app/templates/components/my-component.hbs'}}
  ...
{{/code-snippet}}
```

Similar to how things are set up in the [Ember guides](https://guides.emberjs.com/v2.18.0/components/defining-a-component/).

That much is landed, but the problem I ran into is that this doesn't actually help Markdown code snippets, which we have a ton of e.g. in `quickstart/template.md`. The idea I had was to allow users to do something like this, and translate it into an actual invocation of the `{{docs-snippet}}` component:

```
```js|name='my-snippet'|title='app/controllers/my-controller.js'
```

Does the convenience of being able to do this over just manually invoking `{{docs-snippet}}` in your markdown seem worth the mental overhead of having yet another way to insert a code snippet? Note also that this won't work in API docs, since those are never processed as HBS templates, just straight Markdown --> HTML.

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,8 +72,8 @@ module.exports = {
 
     includer.options.snippetSearchPaths = includer.options.snippetSearchPaths || ['tests/dummy/app'];
     includer.options.snippetRegexes = Object.assign({}, {
-      begin: /{{#(?:docs-snippet|demo.example|demo.live-example)\sname=(?:"|')(\S+)(?:"|')/,
-      end: /{{\/(?:docs-snippet|demo.example|demo.live-example)}}/,
+      begin: /(?:{{#(?:docs-snippet|demo.example|demo.live-example)\s+|```\w+?\|)name=(?:"|')(\S+?)(?:"|')/,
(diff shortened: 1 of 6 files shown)


samselikoff:
Love the idea.

Re: the code snippets within Markdown files, will the Markdown compiler pass-through block-handlebars invocations? If so we could recommend using `{{#docs-snippet}}` directly in .md files

dfreeman:
My leaning at this point is to just switch to using `{{#docs-snippet}}` directly in .md files. I'll close this out and update the quickstart template separately to do that.

PR #47, opened 2018-02-05
Add id to example component

manudwarf:
Hi!

We wanted to be able to link to specific examples in a give page. Would this addition be ok?

### DIFF
--- a/addon/components/docs-demo/x-example/component.js
+++ b/addon/components/docs-demo/x-example/component.js
@@ -6,7 +6,7 @@ export default Component.extend({
 
   classNames: 'docs-demo-example',
 
-  didReceiveAttrs() {
+  init() {
     this._super(...arguments);
     this.set('elementId', 'example-' + this.get('name'));
   }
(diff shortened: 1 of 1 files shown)


samselikoff:
Hm I tend to think of `elementId` as private API to Ember...

@dfreeman thoughts?

manudwarf:
According to the doc, it is public:
https://emberjs.com/api/ember/2.18/classes/Component/properties/elementId?anchor=elementId

dfreeman:
👍  Seems reasonable to me. As a tiny nit, it might be more correct to use `init()` than `didReceiveAttrs()`, since `elementId` becomes read-only after the first render.

samselikoff:
Wow, TIL. Let's change it to `init` and then we can merge away!

Thanks for the contrib!

manudwarf:
Fixed :)

On a separate note, the reason why I'm suggesting this change is because we want live examples to be linkable in our docs.
I will come back with an additional pull request for a linkable anchor. Is there anything in particular I should pay attention to? Do you want such a feature to be configurable, hidden/shown by default?

samselikoff:
Hmm. Now I'm looking at Bootstrap's site and wondering, could we just link to headings?

https://getbootstrap.com/docs/4.0/components/card/#example

That's already on the docket, but I'm wondering if all h1s, h2s and h3s were linkable would it solve your use case? Could you just put heading above each example you wanted to provide a link to?

PR #49, opened 2018-02-06
Easier hbs in markdown

dfreeman:
A few additions/changes here:
 - a preprocessing pass on markdown documents to make them play nicely with handlebars code
 - a switch to stop including extensions in snippet names by default, so we can call them whatever we want independent of the file type they're in, and the declaration always matches the usage
 - super minimal qunit setup with coverage for the markdown preprocessing (will hopefully encourage me to write more node tests in the future)

I _believe_ I have all the snippets updated, but if you happen to have the time/inclination to give this a once over sometime today @samselikoff, a second set of eyes couldn't hurt.


samselikoff:
lgtm!

PR #52, opened 2018-02-14
[refactor] Unify app and addon

pzuraq:
This PR unifies the app and addon folder to make development easier.
Highlights:

* Had to drop pods, no way to control the base app and the way it does
resolution. Also had to drop ember-component-css as a result
* Components are now in /app so needed to set up some basic options for
specifying input paths. Users can add absolute paths or tree names to be
included.
* Customized `treeForAddonStyles` to pull from `app/styles` instead of
`addon/styles`. This allow us to remove addon styles completely, but not
require consumers to manually include SASS files or ember-cli-sass


samselikoff:
can we close?

PR #54, opened 2018-02-16
Auto export addon folder

samselikoff:


index.js
```diff
@@ -205,3 +219,40 @@ class FindDummyAppFiles extends Plugin {
...
+
+      addonFiles.forEach(addonFile => {
+        let module = addonFile.replace('.js', '');
+        let file = path.join(this.outputPath, moduleType, `${module}.js`);
+        ensureDirectoryExistence(file);
+        fs.writeFileSync(file, `export { default } from 'ember-cli-addon-docs/${moduleType}/${module}';`);
+      });
+    });
+
+  }
```
dfreeman: Will this handle `routes`? I saw you removed that from the `app` tree, but I'm not sure I see how that reexport gets generated here.

dfreeman:
Looks like the test failures are caused by an update to the tokenization format used by Marked in the latest release—I just pushed a fix to master that should take care of those.

index.js
samselikoff: Woops! Good catch... some acceptance tests woulda caught that

samselikoff:
Now Ember 2.12 is failing - probably due to this line https://github.com/ember-learn/ember-cli-addon-docs/blob/master/addon/components/docs-viewer/x-main/component.js#L31

PR #55, opened 2018-02-19
[wip] Docs Generator Plugin

pzuraq:
Very very WIP PR, and this really has several different components that should likely be extracted into separate PRs, but pushing now so we can get some visibility on it all:

* Implements a plugin system for API documenters (e.g. YUIDoc, ESDoc, TypeDoc, etc) can plugin into as peer dependencies.
* Rethinks the main documentation models
  * Makes them more JS for focussed and modern, not based on YUIDoc anymore
  * Adds special constructs just for Components
* Converts a bunch of components to ES Classes - NOT really a suggestion to merge, mostly for demoing/figuring out what the data model should be. This could be merged if we like it, but is unrelated to the core work

screen shot 2018-02-19 at 2 58 57 pm


pzuraq:
Closing this in favor of #66, it's no longer needed

PR #57, opened 2018-02-21
Skips non existing paths when exporting addon to app

xcambar:
We have discovered that exporting addon to app breaks
the build when the expected folders do not exist.

This PR simply checks that a folder exists before processing its contents.


samselikoff:
Nice 👍  thanks!

PR #59, opened 2018-02-22
Resolves paths correctly in treeForApp

xcambar:
When building the documentation for an addon with ember-cli-addon-docs,
the paths to ember-cli-addon-docs do not resolve correctly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,12 +117,7 @@ module.exports = {
   treeForApp(app) {
     let trees = [ app ];
 
-    let addonPath;
-    if (this.project.name() === this.name) {
-      addonPath = this.project.root;
-    } else {
-      addonPath = this.project.findAddonByName(this.name).root;
-    }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -117,11 +117,15 @@ module.exports = {
...
     let trees = [ app ];
 
-    let addonPath = path.join(this.project.root, 'addon');
-    let addonTree = new Funnel(addonPath, {
+    let addonPath;
+    if (this.project.name() === this.name) {
+      addonPath = this.project.root;
+    } else {
+      addonPath = this.project.findAddonByName(this.name).root;
+    }
```
dfreeman: I think you can safely just always use the `else` branch here—even in development, `ember-cli-addon-docs` should come up as one of the project's addons

xcambar: Correct. The PR is updated.

dfreeman:
Thanks @xcambar!

PR #60, opened 2018-02-23
Deploy only on master

samselikoff:


samselikoff:
Borrowed the changes from https://github.com/embermap/ember-data-storefront/blob/master/.travis.yml, lmk if they look good.

The floating deps thing, I think is from rwjblue. He said it ensures the addon works on a fresh install.

.travis.yml
```diff
@@ -38,10 +25,42 @@ before_install:
...
+  - versioned tests
+  - deploy
+
+jobs:
+  fail_fast: true
+
+  include:
+    - stage: test
+      env: NAME=test
+      script: yarn test
```
dfreeman: Because the `test` script aliases to `ember try:each`, I think we're running every ember-try scenario in every job: https://travis-ci.org/ember-learn/ember-cli-addon-docs

Do want to make `yarn test` point to `ember try:one $EMBER_TRY_SCENARIO`, or does the Travis config need to change?

.travis.yml
```diff
@@ -38,10 +25,42 @@ before_install:
...
+
+    - stage: additional tests
+      env: NAME=floating dependencies
```
dfreeman: Sorry if I'm missing something obvious, but what does this `NAME` env var do?

.travis.yml
```diff
@@ -38,10 +25,42 @@ before_install:
 install:
   - yarn install --no-lockfile --non-interactive
```
dfreeman: If we're doing a `no-lockfile` install before all our tests run, isn't the default `test` stage already a "floating dependencies" pass?

If I'm understanding the intent right, I think we could probably just remove this global `install` step and let Travis do its default thing (which would be a regular `yarn install`)

.travis.yml
```diff
@@ -38,10 +25,42 @@ before_install:
...
+    - env: EMBER_TRY_SCENARIO=ember-lts-2.16
+    - env: EMBER_TRY_SCENARIO=ember-release
+    - env: EMBER_TRY_SCENARIO=ember-beta
+    - env: EMBER_TRY_SCENARIO=ember-canary
+    - env: EMBER_TRY_SCENARIO=ember-default
 
-after_success:
-  - node_modules/.bin/ember deploy production
+    - stage: deploy
+      if: (branch = master OR tag IS present) AND type = push
```
dfreeman: 👍 I think this is a lot clearer (and based on the weird off-master deploy today, less error prone) than the current setup.

.travis.yml
dfreeman: Looking at ember-data-storefront, I think it has the opposite problem—it looks like it's always running with the default set of dependencies, regardless of the `EMBER_TRY_SCENARIO` variable.

samselikoff: I think we had Storefront copy Ember Data after talking to Robert. I'm poking around Storefront's travis logs but can't see what versions of things it's using. Is there a way to have travis echo out the versions of everything yarn installed?

.travis.yml
samselikoff: I think it how Travis show a description on [this page](https://travis-ci.org/ember-learn/ember-cli-addon-docs)

Here's Robert's commit: https://github.com/emberjs/data/commit/ab4dff563000f1ba9112ec2886cc8c6a8652a803

dfreeman: Ahh, I see. I was only looking at the headings, not the column with the env vars. Does this add much when it's the only job in the stage anyway? ¯\\\_(ツ)_/¯

.travis.yml
dfreeman: I don't know of an easy way to do that, but normally ember-try has some pretty verbose output when it's doing its installation (plus a version grid at the end), and I don't see any of that there.

.travis.yml
samselikoff: Ya Idk, I said the same thing. No idea

samselikoff:
@dfreeman I added back in the global `script` definition. Think this is ready to go?

dfreeman:
Irony. Looks like the most recent build failed because we have a version of console-ui prior to [this fix](https://github.com/ember-cli/console-ui/pull/30) in our lockfile 😬

PR #61, opened 2018-02-26
Versioned deploys

### LINKED ISSUE #2: Versioned guides

### LINKED ISSUE #29: Document deployment in guides
- [x] Out-of-the-box defaults (`gh-pages` branch, hash locationType, etc)
 - [ ] Customizing deploy behavior (hooks in `config/addon-docs.js`)
 - [x] Configuring deployment as part of CI (generating a key pair, setting up GitHub and Travis)

dfreeman:
I'm working through setting up deploys for a fresh addon from scratch as I write out the documentation, so there's still a ways to go there, but wanted to go ahead and get eyes on this. It implements the basic layout we discussed today in Slack:

```
/v1.2.2/...     (deploy of tag v1.2.2)
/v1.2.3/...     (deploy of tag v1.2.3)
/latest/...     (copy of most recent tag build)
/master/...     (deploy of most recent master build)
/versions.json  (list of available versions to power the dropdown)
/index.html     (redirect to /latest, so bare org.github.io/addon-name works)
```

It also updates the `project-version` service to deal with reading the `versions.json` manifest and exposing useful things to do with that information.

Fixes #2, fixes #29


addon/services/project-version.js
```diff
@@ -1,7 +1,35 @@
...
+    this.set('root', rootURL.slice(0, slash));
+
+    if (slash === -1) {
+      this.set('current', 'development');
+      this.set('_versionsPromise', Promise.resolve([
+        { name: 'development', path: '' }
+      ]));
+    } else {
+      this.set('current', rootURL.substring(slash + 1).replace(/\/$/, ''));
+      this.set('_versionsPromise', fetch(`${this.get('root')}/versions.json`)
```
pzuraq: Based on #62 and #64 we should likely update this to use the ajax service

lib/config.js
```diff
@@ -1,15 +1,33 @@
...
+  constructor() {
+    this.repoInfo = gitRepoInfo();
+  }
+
+  shouldUpdateLatest() {
+    if ('ADDON_DOCS_UPDATE_LATEST' in process.env) {
+      return process.env.ADDON_DOCS_UPDATE_LATEST !== 'false';
+    }
+
+    return !!this.repoInfo.tag;
```
dfreeman: Thing that occurred to me as I was writing the documentation: we probably want to change this default behavior not to update `/latest` for prerelease tags. I think `node-semver` has an easy method for checking this on a parsed version string.

dfreeman:
This should be about ready for primetime. The code is in good shape, and I've added a new Deploying section to the guides. I'd love feedback on that if y'all have a chance to take a look—what's unclear, what has too little (or too much) detail, etc.

One missing piece is the ability to handle deploys to custom domains. Right now there are a couple bits that assume the URL will always have a leading segment in the path for the project name. I have a rough idea of what we need to be able to handle that, and I'll either tack it on here or in a follow-on PR, depending on how much of a code change it winds up being.

tests/dummy/app/pods/docs/deploying/template.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Versioning Your Content
+
+Whenever you deploy your documentation site with Addon Docs, it places the compiled application in a subdirectory based on the current state of your git repository. All of this behavior [is customizable](#customizing-deploys), but we expect the out-of-the-box configuration should be a good place to get started.
+
+### Tag Deploys
+
+When you run `ember deploy` at a commit that has a git tag associated with it, the app will wind up in a directory named after that tag. For example, if you've just published version 1.2.3 of your addon (creating tag `v1.2.3` in your git repository), your deployed site will be available at <u>https://**[user]**.github.io/**[repo]**/v1.2.3</u>.
+
+By default, deploying from a tagged commit also places a copy of your app under a special directory called `/latest`. As mentioned above, the `index.html` that Addon Docs sets up at the root redirects to `/latest`, so <u>https://**[user]**.github.io/**[repo]**</u> will always bring developers to the documentation for the most recent stable release of your addon.
```
pzuraq: Does this apply if I'm deploying an update to a previous tag to e.g. fix a typo? Or do we guard against updating latest in that case?

pzuraq:
Overall looks good! Just a couple quick questions, but I think we can merge and move forward

tests/dummy/app/pods/docs/deploying/template.md
dfreeman: Since a typo fix likely wouldn't be from a tagged commit, I think this would be a case where you'd manually do something like this (assuming you're re-publishing `v1.2.3`):

```sh
ADDON_DOCS_UPDATE_LATEST=true ADDON_DOCS_VERSION_PATH=v1.2.3 ember deploy production
```

pzuraq: sounds good! :shipit:

addon/services/project-version.js
```diff
@@ -1,7 +1,36 @@
...
+  current: null,
+  root: null,
 
+  init() {
+    let { rootURL } = getOwner(this).resolveRegistration('config:environment');
+    let slash = rootURL.indexOf('/', 1);
+
+    // TODO deal with apps deployed to custom domains, so their pathnames don't have a leading
+    // segmenet for the project name. This will impact this service and the 404 page.
+    this.set('root', rootURL.slice(0, slash));
```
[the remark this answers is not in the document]
dfreeman: If people use the default layout, we could, but there's technically nothing stopping you from returning something like `version/1.2.3` as the place you want a particular deploy to go, so we can't be certain the version is only a single segment.

That said, that approach actually feels better to me than what's currently here, and I almost built it that way until I realized that the 404 redirect logic _needs_ to know the number of static segments at the start in order to work. Since we can't get away from that, I decided to leave this as is so they could both pull from the same source when we do fix this.

Issue #62, opened 2018-02-26
Fastboot build is not working because of fetch


kolybasov:
### Issue 1

There is fastboot regression introduced in PR [#58](https://github.com/ember-learn/ember-cli-addon-docs/pull/58/files#diff-5e0a877c62054639289d5d5cf6714bc6R23).
```
ReferenceError: fetch is not defined  
```

Possible solution:
Add `ember-fetch` and use `fetch` for making requests:
- [docs-search service](https://github.com/ember-learn/ember-cli-addon-docs/blob/master/addon/services/docs-search.js#L77)
- [application adapter](https://github.com/ember-learn/ember-cli-addon-docs/blob/master/addon/adapters/application.js#L23)

However there is problem with `ember-fetch` itself https://github.com/ember-cli/ember-fetch/issues/62

### Issue 2
One more issue discovered during fastboot tests:

Adding `?fastboot=false`(or basically any query param) to URL throws error because route cannot be found in docs-routes service https://github.com/ember-learn/ember-cli-addon-docs/blob/fb16174de53d5d50e192acb13c0ec29c1f16103d/addon/services/docs-routes.js#L42-L45
This one can be fixed with adding one more replace:
```javascript
.replace(/\?.+$/, ''); // remove query-params
```

pzuraq:
Should be fixed by #64, thanks for the bug report!

PR #64, opened 2018-02-27
[bugfix] Replace fetch with ajax service

pzuraq:
Did a bit of digging into #62 and I think it's going to be easier to use ajax service for the time being than try to solve those issues, so decided to fix it real quick


samselikoff:
👍

PR #65, opened 2018-02-28
[redesign] Adds the new API Pages

pzuraq:
Adds new API pages that are tailored specifically to Ember components
and classes. Highlights include:

* Dedicated Yields and Arguments sections for components
* Import paths for functions and variables in modules
* OOTB module support - no need to use @module!

screen shot 2018-02-27 at 5 04 19 pm
screen shot 2018-02-27 at 5 04 15 pm
screen shot 2018-02-27 at 5 04 07 pm


samselikoff:
looks awesome man! I think we can merge + continue iterating

rwwagner90:
Do these docs include only documentation or will they have demos of the components they document, so you can see the results of the example code?

pzuraq:
Currently only documentation, dynamically parsing docs and rendering demos is tricky, especially when you consider that many demos need to have controller code, router code, etc to function properly. The main demo code is still meant to be done via markdown files and the other components.

Once we get the plugin system in and are iterating, I think we can come back to that and try to experiment some more. It would be nice for simpler components 😄

PR #66, opened 2018-03-01
[feat] Plugin Registry

pzuraq:
This PR pulls out the docs generator into a separate plugin and adds
a plugin registry system based on the one in ec-deploy. It adds the
yuidoc plugin back in by default in the main blueprint, but doesn't
actually include it as a dependency any longer, so existing users will
have to upgrade.

We should bump the minor version after this to indicate it's a breaking change.


blueprints/ember-cli-addon-docs/index.js
```diff
@@ -27,5 +27,17 @@ module.exports = {
...
     this.ui.writeInfoLine('Updated dummy app rootURL for compatibility with GitHub Pages.');
+
+    const hasPlugins = this.project.addons.some(function(addon) {
+      const isPlugin = addon.pkg.keywords.indexOf('ember-cli-addon-docs-plugin') !== -1;
+      const isPluginPack = addon.pkg.keywords.indexOf('ember-cli-addon-docs-plugin-pack') !== -1;
+      return isPlugin || isPluginPack;
+    });
+
+    if (!hasPlugins) {
+      this.addAddonsToProject({
```
dfreeman: Since [`addAddonsToProject` is async](https://ember-cli.com/api/classes/Blueprint.html#method_addAddonsToProject), we probably want to return this promise

lib/models/plugin-registry.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    return this._docsGenerators;
+  }
+
+  _discoverPlugins(addons) {
+    return addons.reduce((plugins, addon) => {
+      if (isPlugin(addon)) {
+        plugins.push(addon);
+      } else if (isPluginPack) {
```
dfreeman: Should this be `isPluginPack(addon)`?

blueprints/ember-cli-addon-docs/index.js
pzuraq: Ah totally, my bad!

lib/models/plugin-registry.js
pzuraq: Yes, yes it should

PR #67, opened 2018-03-02
Speed up Travis

dfreeman:
I took a look at our Travis config, and it turned out we're still running all the ember-try scenarios in each of the first two stages because `yarn test` aliases `ember try:each`. This update:
 - updates the locked/floating dependencies phases to only run the test suite a single time
 - runs the node test suite for each of those (before it was being caught because there's an ember-try scenario for it, but that's not really the ideal way to manage it)
 - removes the `ember-default` try scenario from the versioned tests stage, since that should be equivalent to our 'floating dependencies' run

Before: [~33 minutes](https://travis-ci.org/ember-learn/ember-cli-addon-docs/builds/347354105) of wall clock time
After: [~14 minutes](https://travis-ci.org/ember-learn/ember-cli-addon-docs/builds/348084778) of wall clock time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,8 @@ notifications:
   email: false
 
 stages:
-  - test
-  - additional tests
+  - locked dependencies
+  - floating dependencies
   - versioned tests
   - deploy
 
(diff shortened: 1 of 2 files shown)


samselikoff:
very nice! lgtm

PR #71, opened 2018-03-06
[feat] Adds sandbox and multi-project builds

pzuraq:
Adds a sandbox route to the docs app and a second project to the build,
allowing us to demonstrate different documentors and their techniques.


.npmignore
```diff
@@ -15,6 +15,9 @@ bower.json
 ember-cli-build.js
 testem.js
 
+# sandbox examples
+/lib/sandbox
```
samselikoff: should this whole dir be ignored?

pzuraq: I think so, for the sake of publishing. These sandbox examples should just be available to the dummy app, not anyone who installs, correct?

I don't think this would ignore `/lib`

samselikoff: Oh oh npmignore not gitignore 🤦‍♂️

yep!

lib/sandbox/app/template.hbs/esdoc-component.hbs
```diff
@@ -0,0 +1 @@
+<h2>I'm a esdoc-component</h2>
```
samselikoff: folder name correct?

index.js
```diff
@@ -9,9 +9,12 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app'); // eslint-disable-
...
 const walkSync = require('walk-sync');
 
-const DEFAULT_ADDON_OPTIONS = {
-  includePaths: [],
-  includeTrees: ['addon']
+const DEFAULT_PROJECTS = {
+  main: {
+    tree: null,
+    include: null,
+    includeTrees: ['addon']
```
dfreeman: I'm a little confused by `tree` vs `include` vs `includeTrees` — definitely not a blocker, but it would be great at some point to get some documentation for what each of these is and when you'd use each if it's going to be something we expose.

pzuraq: totally, I'm not sure this is solid yet either so I'm open to suggestions.

* `tree` is an arbitrary tree you can provide. If you do, it'll use that tree instead of anything else.
* `include` is a set of paths that gets passed to a Funnel to include. The reason I went with this over a list of trees is wanting to preserve the actual directory structure - if you pass `'addon'` then that tree will actually move all files/folders in addon to the root of the next input, which is generally not what we want.
* `includeTrees` is a way to add the _trees_ of the parent addon to the same broccoli funnel - this is helpful if users, for instance, change their `app` or `addon` tree for some reason.

I think we can do this in a cleaner way, `includeTrees` may not actually be necessary (how often does someone change their tree path?)

PR #73, opened 2018-03-08
Linkable live docs

manudwarf:
This pull request adds an anchor link.
The hover animation is to make the purpose of the anchor more obvious, otherwise it seemed to have caused some confusion.
I am very open to suggestions on how to make these elements more configurable if you'd prefer.

### DIFF
--- a/addon/components/docs-demo/x-example/styles.scss
+++ b/addon/components/docs-demo/x-example/styles.scss
@@ -1,3 +1,13 @@
 .docs-demo-example {
   padding: 20px;
+  position: relative;
+
+  .docs-demo__anchor {
+    color: black;
+    opacity: 0.5;
+    padding: 10px 15px;
+    position: absolute;
+    right: 0;
+    top: 0;
+  }
 }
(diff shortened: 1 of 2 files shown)


manudwarf:
Any news on this? :)

pzuraq:
There's actually already a style implementation of this in API docs, so I think we should make sure both of them the same. I hacked it out in an afternoon so unsure which solution is better, need to take a look

manudwarf:
Hi,

I wanted to know if anyone has taken a look at this?

Thanks

pzuraq:
Hey @manudwarf, we just merged a large redesign so we haven't had a chance to look at this. It'll likely need to be rebased and combined with the link tags used in API docs already. I'll have a chance to look at it on Friday

samselikoff:
@manudwarf sorry about the delayed response!

If we add a link to demos I'd like to think through the design a bit.

First question, though: if section headings were linkable would that be sufficient for your use case? This is definitely a planned feature and I think is more common in these types of sites than linking directly to the actual snippets.

pzuraq:
Hey @manudwarf, we ended up addressing this and integrating it with the right-hand sidebar nav, you can check it out on the docs site now! Thanks again for suggesting this feature 😄

PR #82, opened 2018-03-12
Fix confusion in the quickstart guide

### LINKED ISSUE #78: Quickstart install issue
`ember install ember-learn/ember-cli-addon-docs`                                                                                                                               

```
Yarn: Installed ember-learn/ember-cli-addon-docs
WARNING: Could not figure out blueprint name from: "ember-learn/ember-cli-addon-docs". Please install the addon blueprint via "ember generate <addon-name>" if necessary.
Installed addon package.
```

`ember s`

```
WARNING: ember-cli-addon-docs needs plugins to generate API documentation. You can install the default with `ember install ember-cli-addon-docs-yuidoc`
Livereload server on http://localhost:7020
Build Error (Babel)

dummy/components/esdoc-component.js: Unexpected token (28:2)

  26 |     The count
  27 |   */
> 28 |   @argument
     |   ^
  29 |   @type('number')
  30 |   count = 0;
  31 |


Stack Trace and Error Report: /var/folders/d2/_qddrkd50vn4wx7vvfxhhtvc0000gn/T/error.dump.57f8d857050249531c059798577b0bb7.log
cleaning up...
````

```
ember-cli: 2.16.2
node: 6.11.4
os: darwin x64
```

alexlafroscia:
Closes #78

### DIFF
--- a/tests/dummy/app/pods/docs/quickstart/template.md
+++ b/tests/dummy/app/pods/docs/quickstart/template.md
@@ -5,15 +5,15 @@ This quickstart guide will get you going with a docs site for your brand new add
 1. **Install the addon.**
 
   ```
-  ember install ember-learn/ember-cli-addon-docs
+  ember install ember-cli-addon-docs
   ```
 
-2. **Add the docs route.** Open `tests/dummy/app/router.js` and add a route named `docs`. Go ahead and make it nested, since this is where you'll be defining additional documentation pages.
+2. **Add the required routes.** Open `tests/dummy/app/router.js` and add the following route definitions:
 
   {{#docs-snippet name='quickstart-router.js' title='tests/dummy/app/router.js'}}
     this.route('docs', function() {
       this.route('api', function() {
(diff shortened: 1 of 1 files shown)


gossi:
While you are at it. In point 4 of the quickstart "Create your docs skeleton". It says the file: `tests/dummy/app/pods/docs/template.hbs` from my understanding should be `tests/dummy/app/templates/docs/template.hbs` (`pods` vs `templates`) - can you confirm my assumption and fix this accordingly?

alexlafroscia:
I think that template is correct... If you have a `docs` pod, then the file name they reference is right.

pzuraq:
We have decided to switch all examples and recommendations to not use `pods` in the docs, because the recommended format by Ember is classic structure. I don't think you need to update that as well (though it would be appreciated) but it seems like a new eslint rule has been added that is causing tests to fail, we need to call `this._super(...arguments)` in an init function somewhere

PR #89, opened 2018-03-13
[bugfix] Ensure sandbox does not get included in parent addons

### LINKED ISSUE #87: CI is broken with projects using ember-cli-addon-docs
Syntax errors are declared starting with this file:

https://github.com/ember-learn/ember-cli-addon-docs/blob/master/lib/sandbox/app/components/esdoc-component.js

The syntax is obviously wrong, but I can't tell whether this was an intentional move, therefore I can't offer a PR or anything.

pzuraq:
Fixes #87


dfreeman:
I think we could probably do this with a little less hacking with `shouldIncludeChildAddon`, something like:

```js
shouldIncludeChildAddon(addon) {
  return addon.name !== 'sandbox';
}
```

pzuraq:
Main concern there is we're going to have to add some logic for checking whether or not we are in our own repo or a consumer, which isn't terrible but tbh I'd rather just have separate concerns entirely. We're using public APIs here, so I don't think it's that hacky

dfreeman:
> we're going to have to add some logic for checking whether or not we are in our own repo

We shouldn't need to care, I think—in our own repo, the `paths` entry will still apply for our dummy app, so everything will still be included, but it will be inert otherwise.

dfreeman:
FWIW I use the same pattern in ember-css-modules, which has several dummy addons for testing different scenarios. https://github.com/salsify/ember-css-modules/blob/master/index.js#L19

pzuraq:
Ah, I see how it works now, both the dummy app and the addon will run to include the in-repo addon.

Ember-CLI is confusing 🙃

PR #96, opened 2018-03-14
[feat] API filtering

pzuraq:
This PR adds filtering of members for classes and components that allows
users to filter out private, protected, inherited, and deprecated
members. In the case of components, users can also filter out `internal`
properties - everything besides Yields and Arguments.

Filters only appear if there is something to filter, and by default
they are enabled (unchecked), so users should see the minimal public API
of the class.

The complexity of the filters is definitely not ideal, but at the moment
I'm unsure if pushing that complexity into the object model (e.g. by
adding `publicFields`, `privateFields`, `staticPublicFields`,
`staticPrivateFields`, etc directly to models) as part of the build step
will help much.

screen shot 2018-03-13 at 5 49 05 pm
screen shot 2018-03-13 at 5 49 00 pm

### DIFF
--- a/addon/components/api/x-class/component.js
+++ b/addon/components/api/x-class/component.js
@@ -1,19 +1,35 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { capitalize } from '@ember/string';
+import { memberFilter }  from '../../../utils/computed';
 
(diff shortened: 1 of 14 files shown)


samselikoff:
killer work

PR #98, opened 2018-03-14
Add `ember-ajax` as a dependency

### LINKED ISSUE #76: Should include `ember-ajax` as a dependency
My addon removed the default `ember-ajax` dependency. `ember-cli-addon-docs` requires it but just assumes it'll be present. `ember-ajax` should be moved to the `dependencies` of the addon (or better yet, removed altogether)

alexlafroscia:
This should avoid addon authors from having to include it in their own dependencies.

Closes #76


pzuraq:
We actually want to switch to `ember-fetch` since ajax won’t work with addons that are trying to remove jquery, was about to open a PR

alexlafroscia:
Honestly, that's a much better idea. As the maintainer, I try to discourage people from using `ember-ajax` whenever possible 😜

PR #99, opened 2018-03-14
[bugfix] Replace ember-ajax with ember-fetch

### LINKED ISSUE #76: Should include `ember-ajax` as a dependency

pzuraq:
Closes #76, supercedes #98


fastboot/initializers/ajax.js
```diff
@@ -0,0 +1,8 @@
+export default {
```
alexlafroscia: Can this be removed altogether?

pzuraq: Tried it out and it seems to work, so we'll roll without it for now

Issue #103, opened 2018-03-15
Add /config/addon-docs.js to npmignore automatically


bravo-kernel:
Any ideas what could be causing this.

When running `yarn run lint:js` in my addon, eslint throws the following error on `/config/addon-docs.js`:

    4:33  error  "ember-cli-addon-docs" is not published  node/no-unpublished-require

- ember-cli: 3.1.0-beta.1
- node: 8.9.0
- os: linux x64    
- ember-cli-addon-docs as a devDependency

> error occured on non-beta as well before

bravo-kernel:
/config/addon-docs.js

```js
/* eslint-env node */
'use strict';

const AddonDocsConfig = require('ember-cli-addon-docs/lib/config');

module.exports = class extends AddonDocsConfig {
  // See https://ember-learn.github.io/ember-cli-addon-docs/latest/docs/deploying
  // for details on configuration you can override here.
};

```

pzuraq:
So this throws when you have a dev dependency and eslint thinks it's _not_ a dev dependency I'm pretty sure, so you'll have to manually mark that file as being dev only

bravo-kernel:
Thanks for replying, I will see if I can somehow mark it. I have absolutely zero JS/Ember skills but I do wonder why this is being done this way (and if I am really the first one encountering this). 

There's a whole thread advertising against it here https://github.com/mysticatea/eslint-plugin-node/issues/47 but if this is the way to go I will try to add a note to the docs.

 Some quotes:

> That rule disallows imports of unpublished files/modules from published files. (Because such imports may cause severe problems after npm publish)

> I would definitely not want to publish code that has require('lodash') if "lodash" is only a devDependency. That would be broken. So I totally appreciate the way this rule currently operates.

bravo-kernel:
@pzuraq not the first I see. Seems mr. @elwayman02 had the same, he simply disabled it.

https://github.com/elwayman02/ember-interactivity/blob/master/config/addon-docs.js

Would this be welcomed as a PR?

ro0gr:
@bravo-kernel you have to add your "/config/addon-docs.js" to the ".npmignore". 

@pzuraq I think it's a job for blueprint?

bravo-kernel:
Works like a charm mr. @ro0gr 👍

bravo-kernel:
Wrong button, totally works. Do you folk want me to close the ticket?

pzuraq:
[...]

PR #105, opened 2018-03-15
Stop publishing "config/addon-docs.js"

ro0gr:
Fixes https://github.com/ember-learn/ember-cli-addon-docs/issues/103


pzuraq:
Thanks for the fix 😄

PR #107, opened 2018-03-15
Remove application adapter and serializer

ef4:
The presence of the application adapter and serializer broke ember-animated's test suite (because I was already using ember-data for another purpose).

In general I think it's risky for an addon to customize the global default application adapter & serializer, even when the consuming app is "just" an addon dummy app.

This PR limits the scope of the adapter and serializer customizations to only the models that are part of ember-cli-addon-docs.


pzuraq:
I want to experiment with pulling the API documentation portion of the docs into an engine so we can get better isolation to prevent things like this from happening, but I think this is the best path forward for now!

ef4:
Yes, engines are a nice possibility for the future, but in their current state I don't think they'll meet your needs.

The biggest gap is that we'd want to allow the host application to inject some components into the engine (or "yield" from the engine back to a host app route). Otherwise you can't easily do fully interactive demos, etc.

pzuraq:
My thinking is that the only portion of the app which would be an engine would be the API docs themselves, which contain all of the models/ember data needs. Basically everything in the `/docs/api` routes, which is not controlled by the user at all at this point

pzuraq:
Thanks for the PR 🎉

dfreeman:
Thanks, @ef4! This had been bugging me for a while but kept getting back-burnered for other things :)

bravo-kernel:
This is a huge fix @ef4, just spent about 6 hours investigating ember-cli-mirage only to end up here 🙄. Does this not warrant a patch-release ?

pzuraq:
It does, there’ve just been quite a few things that went in/are going in so I wanted to wait until I had a chance to fully review tomorrow. Will see if I can get to it today

PR #112, opened 2018-03-16
Add qunit-dom

cibernox:


pzuraq:
It seems like this would be better handled with page objects, which we’ve already started to adopt (see the other integration tests)

ro0gr:
@pzuraq unfortunately I can't agree. `qunit-dom` currently is more expressive here:
```
assert.dom('h1').hasText('DocsHero');
```

pzuraq:
@ro0gr but that would only be because we haven’t yet setup a page object for that page, correct? Again see the other tests where we currently make similar assertions.

My main concern is that we do things consistently here. If PageObjects are not the way to go, that’s fine too, but we should not be writing tests differently in different circumstances just because.

cibernox:
I am not a fan of page objects myself, but whatever the majority prefers.

That said, although I _can_ see the point of ember-page-objects for acceptance tests, for _component integration_ tests I think that `qunit-dom` is far more convenient

ro0gr:
Agree about consistency. In the long run page-objects are less brittle I believe.

ro0gr:
@cibernox one of the page-object strengths is that you can reuse it in both kinds of tests

pzuraq:
I think the boilerplate surrounding page object in component-integration tests needs to get better, we’re working on that story at Addepar. 

Main benefits come from composeability and reuse - when composing a component from many other components, writing test helpers is as simple as combining the individual component-page-objects.

ro0gr:
> I think the boilerplate surrounding page object in component-integration tests needs to get better

actually you don't need `page.setupContext()` API with the new tests API

pzuraq:
@ro0gr that's true, but there is still a fair amount of boilerplate you have to do to get the level of testing you want in some cases. I've experienced this in the [tests for our popover components](https://github.com/Addepar/ice-pop/blob/master/tests/integration/components/ice-dropdown-test.js#L230-L241), I haven't been able to pinpoint exactly what it is about them but it feels like it could be simplified, definitely something that we can improve in the page object testing story.

Based on the convo in slack, I think it's reasonable to use `qunit-dom` for integration tests. We can merge this PR for now and then convert it to Page Objects when we get a chance

ro0gr:
@pzuraq yes, there is still an area for PO's API improvements

pzuraq:
Going to merge this for now and we can come back to page-ify those acceptance tests when we get time. Thanks for the help @cibernox!

Issue #68, opened 2018-03-04
Update `demoURL` on installation


dfreeman:
Ember Observer looks for a `demoURL` key under `ember-addon` in an addon's `package.json` to optionally add a link to that addon's demo/docs site. We could set this for the host in our default blueprint if there isn't one already present.

kategengler:
I think using the "homepage" field in package.json should also be considered. It is used by the `npm docs` command to open a doc site. For example, if set on ember-try `npm docs ember-try` would open the correct website.

spencer516:
@kategengler — anything I can do to help getting Ember Observer support for the `homepage` property? I'm happy to put together a PR — but it's not clear where the Ember Observer API code is...?

(I've just updated the PR to add this functionality and use `homepage`: https://github.com/ember-learn/ember-cli-addon-docs/pull/114 ).

dfreeman:
Thanks @spencer516!

PR #114, opened 2018-03-17
Add `demoUrl` to add-on config in default blueprint

spencer516:
Adds feature described in https://github.com/ember-learn/ember-cli-addon-docs/issues/68. 

The approach here follows these steps:

- If there is already a `demoUrl`, don't do anything. 
- Attempt to determine the Github repository URL
  - First check `package.json`
  - Fallback to the _first_ remote host in the git config 
- Attempt to parse the Github URL and set `demoUrl` to `https://${gitInfo.user}.github.io/${gitInfo.project}`
- If unable to find repo or unable to parse it, provide warning to the console with instructions to manually setup the `demoUrl`


dfreeman:
@spencer516 Overall this looks great! I think there are just a couple lint errors in the node code.

@kategengler pointed out over in #68 that the `homepage` field in `package.json` might be more broadly applicable. If EmberObserver were to start displaying that field somewhere, my leaning would be to say we should use it and just apply exactly the logic you've built here to that field instead of `ember-addon.demoUrl`.

@samselikoff @pzuraq thoughts?

spencer516:
@dfreeman — awesome. I'll address lint errors and await some additional thoughts about the `homepage` vs `ember-addon.demoUrl`. (Adopting the npm "standard" seems 👍 )

pzuraq:
I prefer using homepage as well, more tools that are standardized on the same pattern the better 😄

spencer516:
Just updated to use `homepage` and to address them linting errors.

samselikoff:
lgtm! Awesome PR spencer... thank you !

Issue #34, opened 2017-12-22
Add info about router-scroll configuration


samselikoff:

dfreeman:
Does this just mean recommending using `ember-router-scroll`? The 404 redirect shenanigans to avoid needing to use `locationType: 'hash'` should all be automated now.

samselikoff:
A bit more - to use `ember-router-scroll` with our `{{docs-viewer}}` component, you need to [just a bit of wiring](https://github.com/ember-learn/ember-cli-addon-docs/blob/master/tests/dummy/config/environment.js#L11-L13):

```
routerScroll: {
  scrollElement: '#docs-viewer__scroll-body'
}
```

samselikoff:
This might change after the redesign... I'd like to make the `<body>` always be the main scrolling element

samselikoff:
Not needed anymore

PR #116, opened 2018-03-17
Full rewrite of the Quickstart Guide

bravo-kernel:
Hope you folks appreciate this rewrite of the Quickstart Guide:

- [x] now actually leads to the expected result (working doc site if steps are followed)
- [x] more novice friendly (no prior knowledge required)
- [x] optional configuration moved to separate part (at the bottom) for easy extending the guide without overhauling the main instructions
- [x] removes last pod reference
- [x] adds an additional step for the `addon marketing homepage`

There are three things on my list that need checking.

- [x] ~is it correct that API Reference (and subitems) are auto-generated?~
- [x] ~I found the use of `templates` all over the place very confusing. Not in the Ember context but when looking for the actual documentation (`.md`) files. Perhaps it could be an improvement if we come up with a good name to make the distinguishment. I threw in `Documentation Sources` for now.~
- [x] ~I need some input to describe what's needed for `Ember Router Scroll`. Did find this but was unsure~. Probably better to await the redesign as mentioned in ticket https://github.com/ember-learn/ember-cli-addon-docs/issues/34


alexlafroscia:
In the spirit of being a "quickstart" guide, maybe the `Installation` page setup isn't that necessary? Or maybe the example could have a `Usage` page instead? For most addons, installation is just

```
ember install name_of_addon
```

so suggesting a page in the documentation to cover those three words doesn't feel impactful. Otherwise, I think these changes look great!

alexlafroscia:
Sure thing. Personally think a basic "this is how you use the addon" (outside of API documentation) is more useful than installation steps. I think addon authors will know if their installation is complex enough to warrant a page dedicated to it!

bravo-kernel:
@alexlafroscia getting the quickstart working, solid and result-guaranteed is the first step IMO. Even if it's only to prevent the shame of having a documentation addon with a failing first-encounter documentation page 😉

pzuraq:
Awesome work @bravo-kernel 😄 This is good stuff. 

I'm not super sure about calling templates "documentation sources", I think we may want to work on our fundamentals here. One of the main points of `addon-docs` is that your addon docs are Just An Ember App - just like the Dummy app is today. We may want to figure out how to better setup and convey that point to users so that they understand that all `addon-docs` is, in the end, is:

1. A markdown template preprocessor
2. A bunch of helpful components
3. An opinionated layout (navbar and stuff)
4. Autogenerated API docs

I think 4 is really the trickiest part because of the amount of custom things we enforce (like adding mandatory routes to the router). I have some ideas for how to separate that out a bit more, will see if I can spike something out soon. Would like to see what @samselikoff and @dfreeman think about the wording change.

In the end, definitely on board with this PR in general, thanks so much!

bravo-kernel:
@pzuraq great feedback. I intended this to be a draft so no issue whatsoever to change the wording. The reason I did not choose `Markdown templates` was because I saw a ticket or slack mentioning that other formats could be used as well. 

If.. markdown is the only source for the "static" docs I think `Markdown templates` would make a whole lot of sense. Would have definitely helped me when I first looked.

BTW I think your four points would do very well on the current "Introduction" page, that would have clarified a lot to me as well.

bravo-kernel:
Latest commit:

- uses `Markdown templates` as wording again
- replaces the `index` copy-paste template with a better example using some of the components (very cool stuff folks!)

Feel free to squash and merge.

tests/dummy/app/pods/docs/quickstart/template.md
```diff
@@ -42,15 +46,48 @@ This quickstart guide will get you going with a docs site for your brand new add
     {{/docs-viewer}}
   {{/docs-snippet}}
 
-5. **Fill out the index of your docs page.** We called it Overview but you can call it whatever you want. Since Addon Docs supports markdown templates out of the box, we can make this a Markdown file.
+4. **Create your Markdown templates.** Markdown templates contain the actual
+actual documentation for your addon and live in the folder
```
pzuraq: typo, `actual actual`

tests/dummy/app/pods/docs/quickstart/template.md
```diff
@@ -63,28 +100,30 @@ This quickstart guide will get you going with a docs site for your brand new add
...
+7. **Launch your docs site**. Run `ember serve` and browse to
+`localhost:4200/docs` to enjoy your brand new documentation website.
 
-  {{#docs-snippet name='quickstart-more-docs.js' title='tests/dummy/app/router.js'}}
-    this.route('docs', function() {
-      this.route('installation');
-    });
-  {{/docs-snippet}}
+8. **Create more pages.** To add more doc pages simply follow the same steps as
+used for the `Usage` page in above instructions:
```
pzuraq: probably should be `page in the above instructions`

bravo-kernel: @pzuraq How about `Usage page in above examples` ?

This way users can reverse engineer/lookup what's needed to create an extra docs page.

tests/dummy/app/pods/docs/quickstart/template.md
```diff
@@ -63,28 +100,30 @@ This quickstart guide will get you going with a docs site for your brand new add
...
+
+  {{docs-snippet name="docs-demo-application-template.hbs" title="tests/dummy/app/templates/application.hbs"}}
 
-8. **Round out your site.**
+2. **Optional: add a favicon.** You may want to place a favicon in the
+`tests/dummy/public` folder. We recommend using using
+ [Ember CLI Favicon](https://github.com/davewasmer/ember-cli-favicon).
 
-* **Add a favicon to your dummy app.** We recommend using [Ember CLI Favicon](https://github.com/davewasmer/ember-cli-favicon).
+3. **Optional: better scrolling.** You may want to install
```
dfreeman: Do these need the `Optional:` prefix since they're already under an `## Optional` heading?

bravo-kernel: No they don't. I just forgot to remove those after I did the first one. Nice catch, just pushed an update.

Issue #46, opened 2018-01-27
Build failure with scoped package addon


mriska:
Hi, I was trying out this addon in one of our internal addons. Our addon is hosted privately in an organization scope on npmjs and the `name` in `package.json` is in the form `@organization/package-name`.

When running `ember serve` I got the following error message: 

```shell
› ember s
Livereload server on http://localhost:7020
The Broccoli Plugin: [BroccoliMergeTrees] failed with:
Error: ENOENT: no such file or directory, open '/Users/xxx/work/package-name/tmp/docs_generator-output_path-Krhjlfrc.tmp/docs/project-versions/@organization/package-name-0.5.1+344612b9.json'
  at Object.fs.openSync (fs.js:646:18)
  at Object.fs.writeFileSync (fs.js:1291:33)
  at Object.writeFileSync (/Users/xxx/work/package-name/node_modules/ember-cli-addon-docs/node_modules/jsonfile/index.js:117:13)
  at DocsGenerator.build (/Users/xxx/work/package-name/node_modules/ember-cli-addon-docs/lib/broccoli/docs-generator.js:39:8)
```

And when checking the build output folder `docs_generator-output_path-Krhjlfrc.tmp/docs` I could see that the folders `project-versions` and `projects` were empty. 

When temporary removing the scope from the package name in `package.json` the project builds as expected. 

Should the folder name maybe be taken from the `name` specified in `index.js` instead?

dfreeman:
@mriska I have some code sitting on a local branch called `i-had-these-pending-changes-but-im-not-sure-what-they-were-doing` and seeing this issue reminded me that _that's_ what that code was for! I need to make sure it's actually complete, but I've definitely run into this issue as well and will hopefully have it fixed soon.

I agree that we should treat the name in `index.js` as the definitive source of truth, since that's how Ember CLI sees it, and that's where the base path for module imports comes from as well.

mriska:
Great, I actually started something similar myself to test it out, but didn't want to turn it into a PR before I have heard from you that you think it is actually the right approach.

mriska:
@dfreeman Do you want me to help out with fixing this one or do you have a fix that is close to finished?

dfreeman:
[...]

PR #117, opened 2018-03-17
support for scoped packages

### LINKED ISSUE #111: DocsLogo component causes page fail for non-ember starting addon names
Any idea what could be causing this? Fresh addon with most basic Addon Docs, have not added the component in any of my pages.

> Assertion Failed: You passed "undefined" to the docs-logo component, but the only valid options are [ember,ember-cli,ember-data]."

Dhaulagiri:
* Ensures that we can correctly create a nested directory in the event that a package's name is `@myscope/package-name`.  This was breaking because the `@myscope` directory wasn't being created before.
* Ensures that we don't rely on matching names to be at the start of a package's name to allow for a scope to be at the beginning

Fixes https://github.com/ember-learn/ember-cli-addon-docs/issues/46, but I'm honestly not 100% sure if this is the best approach.
Fixes #111

### DIFF
--- a/addon/components/docs-viewer/x-nav/component.js
+++ b/addon/components/docs-viewer/x-nav/component.js
@@ -20,11 +20,11 @@ export default Component.extend({
   addonLogo: computed(function() {
     let name = packageJson.name;
     let logo;
-    if (name.match(/^ember-cli/)) {
+    if (name.match(/ember-cli/)) {
       logo = 'ember-cli';
-    } else if (name.match(/^ember-data/)) {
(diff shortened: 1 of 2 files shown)


pzuraq:
Based on what’s going on in #111 I think I’d like to reduce the logo constraint even further. It should probably default to `ember` if nothing else is matched, and should have a manual override

bravo-kernel:
Defaulting to the `Ember` logo would be awesome, much appreciated

pzuraq:
Awesome, thanks for taking this on!

PR #127, opened 2018-03-26
Clean up `logo` code

alexlafroscia:
* Used in two places, on of which was broken. Split out into a separate function that both locations can import
* Only the `docs-header` used the `addonLogo` property, all other references to the same feature use `logo`. The documentation also used `logo` but the implementation was setting `addonLogo` and ignoring the provided `logo` property. Now it's all just `logo`

### DIFF
--- a/addon/components/docs-header/component.js
+++ b/addon/components/docs-header/component.js
@@ -33,7 +33,7 @@ export default Component.extend({
 
   packageJson: packageJson,
 
-  addonLogo: addonLogo(packageJson),
+  logo: addonLogo(packageJson),
 
   addonName: computed(function() {
     let name = packageJson.name;
--- a/addon/components/docs-header/component.js
+++ b/addon/components/docs-header/component.js
@@ -35,7 +35,7 @@ export default Component.extend({
 
   logo: addonLogo(packageJson),
 
-  addonName: computed(function() {
+  name: computed(function() {
     let name = packageJson.name;
     name = name.replace('ember-data-', '');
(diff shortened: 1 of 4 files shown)


addon/components/docs-header/component.js
```diff
@@ -32,20 +33,7 @@ export default Component.extend({
...
 
-  addonLogo: computed(function() {
-    let name = packageJson.name;
-    let logo;
-
-    if (name.match('ember-cli-')) {
-      logo = 'ember-cli';
-    } else if (name.match('ember-data-')) {
-      logo = 'ember-data';
-    } else if (name.match('ember-data-')) {
```
alexlafroscia: This was using `ember-data-` twice which didn't make sense

addon/components/docs-viewer/x-nav/component.js
```diff
@@ -17,19 +18,7 @@ export default Component.extend({
...
-      logo = 'ember-cli';
-    } else if (name.match(/ember-data/)) {
-      logo = 'ember-data';
-    } else {
-      logo = 'ember';
-    }
-
-    return logo;
-  }),
+  addonLogo: addonLogo(packageJson),
```
alexlafroscia: Unsure if you prefer passing `packageJson` into the CP macro or would rather `this.get('packageJson')` within the CP definition. I like this API better since it makes less assumptions about that `packgeJson` property being available.

samselikoff:
This is great - thank you dude!

PR #130, opened 2018-03-27
Fix template edit link

alexlafroscia:
- More generic matching (removes assumption that the file is within the `pods` directory)
- Fixes some references to the original `docs.api.class` route and updates them to use `docs.api.item`

### DIFF
--- a/addon/components/docs-viewer/x-main/component.js
+++ b/addon/components/docs-viewer/x-main/component.js
@@ -35,8 +35,9 @@ export default Component.extend({
 
       return `${packageJson.repository}/edit/master/addon/components/${klass}/component.js`;
     } else {
-      let templatePath = `pods/${path}`;
-      let file = appFiles.find(file => file.match(`${templatePath}/template.(hbs|md)`));
+      let file = appFiles
+        .filter(file => file.match(/template.(hbs|md)/))
+        .find(file => file.match(path));
 
       return `${packageJson.repository}/edit/master/tests/dummy/app/${file}`;
     }
--- a/addon/components/docs-viewer/x-main/component.js
+++ b/addon/components/docs-viewer/x-main/component.js
@@ -3,7 +3,7 @@ import Component from '@ember/component';
 import layout from './template';
 import { computed } from '@ember/object';
 import appFiles from 'ember-cli-addon-docs/app-files';
-import { dasherize } from '@ember/string';
+import addonFiles from 'ember-cli-addon-docs/addon-files';
 import config from 'dummy/config/environment';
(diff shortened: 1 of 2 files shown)


addon/components/docs-viewer/x-main/component.js
```diff
@@ -23,20 +23,23 @@ export default Component.extend({
...
       return;
     }
+
     path = path.replace(/\./g, '/');
 
-    if (path === 'docs/api/class') {
-      let params = getOwner(this).lookup('route:application').paramsFor('docs.api.class');
-      let klass = dasherize(params.class_id.replace(/-.+$/g, ''));
-      let path = `pods/${path}`;
+    if (path === 'docs/api/item') {
```
alexlafroscia: `docs/api/class` was what the original documentation had, but no longer matches what the guide suggests.

addon/components/docs-viewer/x-main/component.js
```diff
@@ -3,7 +3,7 @@ import Component from '@ember/component';
 import layout from './template';
 import { computed } from '@ember/object';
 import appFiles from 'ember-cli-addon-docs/app-files';
-import { dasherize } from '@ember/string';
+import addonFiles from 'ember-cli-addon-docs/addon-files';
```
alexlafroscia: Needed to add the logic to also expose the `addon` tree contents, otherwise you had to really guess a lot about what the current URL on Github would be. At least now we have a reference of the actual files to try and match against.

addon/components/docs-viewer/x-main/component.js
```diff
@@ -23,20 +23,23 @@ export default Component.extend({
...
-      // `routing` doesn't exist for old ember versions via ember-try
+      // `router` doesn't exist for old ember versions via ember-try
       return;
     }
+
     path = path.replace(/\./g, '/');
 
-    if (path === 'docs/api/class') {
-      let params = getOwner(this).lookup('route:application').paramsFor('docs.api.class');
-      let klass = dasherize(params.class_id.replace(/-.+$/g, ''));
```
alexlafroscia: `class_id` is no longer what the catch-all param for the route is called.

addon/components/docs-viewer/x-main/component.js
```diff
@@ -23,20 +23,23 @@ export default Component.extend({
...
-      let path = `pods/${path}`;
```
alexlafroscia: As before, remove the assumption that the files are within `pods`

alexlafroscia:
Not sure if these changes will be compatible with _every_ addon, but it's a step toward the right direction I think.

PR #139, opened 2018-04-07
[api] Remove helper items, rework navigation index

### LINKED ISSUE #136: Helpers are not discovered
I am struggling to get helpers section to display. 

```
{ file: 'addon/helpers/my-helper',
  line: 3,
  hide: '',
  class: 'addon/services/file-queue~FileQueue',
  module: undefined }
```

I just added few console.logs to lib/models/function.js on yuidoc package. It seems it matches helper to random service. That explains why I saw part of docblock from helper file in my service methods list. Helper’s docblock had @method defined.

pzuraq:
It turns out that helpers can actually be functions _or_ classes, which
was not something which the original data model took into account.
This PR removes the notion of a unique "helper" type from the data
model in favor of deferring to whatever construct is ultimately
resolved: Either a class or a function.

The navigation index code has also been simplified a bit - it was pretty
hacked together after EmberConf, so it needed some straightening out.
For now, the rule is that every resolvable file gets an entry, and the
the last segment of the file path becomes the entry's name. Helpers and
components are {{curlied}} and everything else is CapitalCased.

We are also no longer making assumptions about what a file is exporting.
This means that if  helper function or class is exported from resolved
directory, it will _not show up_ in the navigation at all. We can work
on solving this use case later on, but I think it will be sufficiently
rare that we can punt on it until later. `utils/` folders are
purposefully excluded from this and are placed instead in the module
structure, since that is the most common directory for helper functions.

We also need to consider how this structure may change with module
unification and nesting, but that's a problem for the future.

Closes #136


pzuraq:
Failures are due to an unrelated bug upstream that has been fixed. The fix has not yet propagated to Beta, but everything is green on Canary so going to merge this and publish for now.

Issue #149, opened 2018-04-14
"Multiple modules match the path" assert triggered for common name prefix


allthesignals:
Our [addon](https://github.com/kturney/ember-mapbox-gl) namespaces modules with a prefix like this:

mapbox-gl
mapbox-gl-layer
mapbox-gl-call
etc...

Navigating to the generated `mapbox-gl` route throws:
Error:
> "Assertion Failed: multiple modules match the path 'components/mapbox-gl', ids: addon/components/mapbox-gl, addon/components/mapbox-gl-call, addon/components/mapbox-gl-control, addon/components/mapbox-gl-image, addon/components/mapbox-gl-layer, addon/components/mapbox-gl-marker, addon/components/mapbox-gl-on, addon/components/mapbox-gl-popup, addon/components/mapbox-gl-source"

[Source](https://github.com/ember-learn/ember-cli-addon-docs/blob/f032d9add87bd0084de580315e83f7292b4b7030/addon/routes/docs/api/item.js#L24)

Is there a workaround for this? Is matching done with regular expressions or exact match?

allthesignals:
Right now, I'm just overriding the route's model hook and removing the asserts, which is unsafe. Using a last path segment (`const lastPathSegment = path.split('/').pop();`) might work, but could create naming collisions because it strips out the path. 

The `file` property on the model is useful, but it includes a prefixed `/` so you can't use it as a lookup without some formatting. Not sure which solution path is going to break other scenarios...

pzuraq:
I think that logic is faulty, we should be able to match on exact matches I'm pretty sure. I need to think about why we didn't do that in the first place, but it seems broken. This is a duplicate of #94 so I'm going to close for now, we are tracking and will try to get a fix out soon

allthesignals:
@pzuraq alright, I have a branch for this, can push up later and see what you think

Issue #94, opened 2018-03-13
Issues with nested components with similar names


alexlafroscia:
I have a set of two components, `step-manager` and `step-manager/step` (which is `yield`-ed by the manager). I'm running into a number of issues with this scenario

- [ ] I can't generate a documentation page for the `step-manager/step` component because I get an error about the name already being used

<img width="879" alt="screen shot 2018-03-13 at 2 42 01 pm" src="https://user-images.githubusercontent.com/1645881/37371483-b92e9262-26cc-11e8-9231-100cbecc3036.png">

- [ ] The properties on the `step-manager/step` component appear on the `step-manager` page when turning documentation for `step-manager/step` off

pzuraq:
Are you using YUIDoc or ESDoc? What does your documentation look like? Trying to figure out if this is just an app problem or if it's also a build time problem

alexlafroscia:
I pushed my WIP documentation branch here:

https://github.com/alexlafroscia/ember-steps/tree/87a69c5e516eae115f6f48396e4a167bb4eef8c8/addon/components

The `step-manager.js` and `step-component/step.js` are the files in question

alexlafroscia:
It's possible that I accidentally used ESDoc when I should have used YUIDoc... I didn't even realize that was "a thing" that I needed to think about.  The documentation that I'm trying to generate pages for was written a long time ago

pzuraq:
Ah, so you are using YUIDoc (it's the default that gets installed, you can see it in `package.json`). It looks like this issue is that you haven't properly labelled the `step-manager/step.js` component as a class. YUIDoc is very dumb, it does not parse your code at all, so you must tag basically everything. ESDoc is much more intelligent about this, but can only be used with native classes.

pzuraq:
You can read more about how to use the YUIDoc plugin on it's [README](https://github.com/ember-learn/ember-cli-addon-docs-yuidoc#usage), we're working on adding examples to the Sandbox so people can use those as "guides" 😄

alexlafroscia:
Right, so... If I _do_ annotate `step-manager/step.js` as a class, I end up getting an error from the Ember router (see the screenshot in the description of the issue).

pzuraq:
Ok, cool, so two separate issues.

[...]

PR #150, opened 2018-04-15
Changes the way lookups are performed, using file path

allthesignals:
Addresses #94, this does an exact lookup of the module by `file` property, along with some formatting (rather than id). This allows similar-named modules (components specifically) to be listed without asserted errors about multiple modules. 

@pzuraq mentioned we need to look into why the model does the lookup this way currently (using String#match, see #149).

### DIFF
--- a/addon/routes/docs/api/item.js
+++ b/addon/routes/docs/api/item.js
@@ -17,7 +17,7 @@ export default Route.extend({
         || module;
     } else {
       let modules = this.store.peekAll('module');
-      let matches = modules.filter(m => m.id.match(path));
+      let matches = modules.filter(m => (m.get('file') === `/${path}`));
       let module = matches[0];
 
       assert(`no modules match the path '${path}'`, matches.length > 0);


mkay581:
Thanks @allthesignals! Any word on getting this in anytime soon?

allthesignals:
@mkay581 hmm let me see why the test is failing. I think @pzuraq wanted to think about it some more, too... if this is blocking you, you can just overwrite it in your project.

pzuraq:
So I’m planning on reworking the index itself, along with the parhs for modules in general, tomorrow. When I got started on working on the docs for `ember-decorators` over the weekend I realized there were a lot of issues with the way we hacked things together for EmberConf 😅

Still thinking everything over, I want to make sure we nail it this time. Will write up a proposal tomorrow

dfreeman:
@pzuraq As you're thinking through that, coming up with a principled approach to URLs that handles both top-level items (modules/classes/etc) and their individual members (functions/methods/etc) would be fantastic.

I started in on fixing the `SearchIndexer` to include API docs today and realized that while you definitely want to show `MyClass` if you type `myc`, it would also be really nice to show that class's `doSomething()` method in the results and link directly to it if you type `dosom`. The same goes for being able to search exported functions and constants from modules, etc.

pzuraq:
Closing this as it was superceded by #167 and #166

PR #152, opened 2018-04-18
fix(quickstart): index header should match the nav name

knownasilya:
It's called 'Introduction' in the nav block above, so should probably match here as well.


samselikoff:
👍

PR #160, opened 2018-04-29
Fail gracefully in the version selector instead of requiring Mirage.

dfreeman:
How's this look @samselikoff?


addon/services/project-version.js
```diff
@@ -19,7 +19,7 @@ export default Service.extend({
     this.set('current', currentFromURL || 'latest'); // dev-time guard. Think of a better way?
 
     let response = yield fetch(`${this.get('root')}/versions.json`);
-    let json = yield response.json();
+    let json = yield response.ok ? response.json() : [{ name: 'latest', sha: 'abcde', path: '/' }];
```
samselikoff: let's use `"tag": packageJson.version` instead of `sha` here so developers installing/using AddonDocs see their current version in the header

PR #162, opened 2018-04-29
Move deploys under a version namespace

dfreeman:
Supporting root deploys looks like it'll only be a small piece of work on top of this, so I'm inclined to say we should land that as well (in a separate PR) to help sweeten the bitter pill of this migration...

@samselikoff here are my notes from migrating my deploy playground addon over, if you get the chance to give this a try with storefront at some point:

- checkout gh-pages (or go to the worktree directory)
- ~open 404.html and change `segmentCount` from 2 to 3~ no longer necessary now that this depends on the `configurable-root-url` branch
- open `index.html` and change the redirect to `./v/latest`
- create a directory called `v` and move each deployed tag/branch directory inside (strip the leading `v` off tag directories)
- for each deployed tag/branch, go open index.html and find/replace `<project>/<branch-or-tag>` with `<project>/v/<branch-or-tag>` (strip the leading `v` off any tag names)
- in `versions.json`, update the `path` for each entry to start with `v/` (strip the leading `v` off any tag paths)
- you should be left with the following in the root (plus possibly some gitignored things like `dist` and `node_modules`)
  - index.html
  - 404.html
  - versions.json
  - v/

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -22,7 +22,8 @@ module.exports = class AddonDocsConfig {
     }
 
     if (this.repoInfo.tag) {
-      return this.repoInfo.tag;
+      // Turn v1.0.0 -> 1.0.0 since we're prefixing with /v/
+      return this.repoInfo.tag.replace(/^v(\d)/i, '$1');
     }
 
     return this.repoInfo.branch || process.env.TRAVIS_BRANCH;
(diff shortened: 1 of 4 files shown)


dfreeman:
Note: this now depends on #163, which we might want to merge and release with the current stuff on master as `0.4.2` before the breaking stuff in here lands and pushes us to `0.5.0`.

PR #164, opened 2018-04-29
Deploy 'latest' into the root of gh-pages

dfreeman:
This builds on the changes in #163 and #162 to allow us to deploy the latest stable release to the root of the destination branch rather than `/latest`. Anyone who wants to try this out should first follow the directions in #162 to migrate any existing deploys over to the new namespaced structure.


dfreeman:
Superseded by #196

PR #166, opened 2018-04-30
[refactor] Simplify and generalize Module links

pzuraq:
This changes the way that we link to modules in order to make it more
generic. The main difference here is that we no longer assume that the
project name has any relation to the module itself, instead we always
use the fully-qualified module name as its id, its link, and title.

There are some pros and cons to this approach:

* Pros
  * More inline with JS/modules as a whole - Since we base everything on
  the full module name, we aren't basing it on arbitrary Ember-only
  concepts, like the Addon/App folders, etc. The build system now just
  receives a tree with a module structure, and assumes that it is the
  correct, canonical structure.
  * More self-documenting - users can know at a glance what the import
  path for an item is just by looking at the sidebar
  * Supports documenting multiple addons in a single project - this is
  important for libraries like ember-decorators that are made up of many
  small packages

* Cons
  * Slightly more verbose - the name of the addon is generally included
  in titles, urls, and import paths, which can be redundant in the
  simple case. This could be special-cased in the future, so we at least
  don't include it in the title for the navigation index.
  * Could be problematic with Node/Browser overlap - users could try to
  have two sets of modules with identical import paths for both the
  Browser and Node (e.g. both have an `index.js`). This is problematic
  in general though, and moving forward Ember-CLI seems to be on a path
  to making sure addons don't do this because its confusing.

Other changes:

* Instead of nesting modules under a `/root` url, we're now nesting them
under `/modules`, so we don't imply that it has anything to do with the
structure of the addon. Everything is just a JS module.
* Fixed routing to resolvable items, components with similar names
should now work.
* Removed module hoisting for now - it's fairly complex, and until we get
all of the modules design and UX figured out I'd rather not be working with
it as a constraint.

### DIFF
--- a/addon/components/api/x-import-path/component.js
+++ b/addon/components/api/x-import-path/component.js
@@ -1,11 +1,7 @@
 import Component from '@ember/component';
 import layout from './template';
-import config from 'dummy/config/environment';
-
-const projectName = config['ember-cli-addon-docs'].projectName;
 
 export default Component.extend({
   layout,
-  classNames: ['import-path'],
-  projectName
+  classNames: ['import-path']
 });
--- a/addon/components/api/x-import-path/template.hbs
+++ b/addon/components/api/x-import-path/template.hbs
@@ -7,4 +7,4 @@
 {{/if}}
 
 <span class="hljs-keyword">from</span>
-<span class="hljs-string">'{{projectName}}{{item.file}}'</span>;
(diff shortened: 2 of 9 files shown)


pzuraq:
Yeah, my plan is to fix that in a followup PR, didn't want to do too much at once. Should be able to get it up sometime this week!

PR #169, opened 2018-05-02
update ember-cli-tailwind

miguelcobain:
This should remove an ember-cli-babel related deprecation when building projects.


pzuraq:
Updated this in a separate PR, and added a caret so we don't have this issue in the future. Thanks @miguelcobain!

PR #173, opened 2018-05-05
[feat] General On This Page Index

pzuraq:
Adds a general On This Page index which looks at the `main` section on
the page to find any `h2`/`h3` elements and create a table of contents
out of them. Notes:

* H1 was omitted because it would be redundant, H1 is semantically meant
to be the title of the page
* Uses a MutationObserver to update the index when changes occur that
weren't caused by routing. This happens in API docs with the toggles.

May need some light touch up, I'm in a rush today 😄


samselikoff:
Took a page from [Read the docs](https://redux.js.org/) styling and went with 
 - a light border
 - one text size
 - one margin for all items
 - a deeper indent for sub items

image

pzuraq:
Looks good! Like the changes 😄

PR #176, opened 2018-05-06
[style] Cleans up the API docs

pzuraq:
Cleans up the API docs visually by putting all metadata in a
single panel at the end of each items' section

screen shot 2018-05-06 at 6 26 58 am


pzuraq:
@samselikoff I'm going to merge this for now so I can move forward with the `ember-decorators` docs, but would like your feedback on the design of the API docs page at some point. I think it'll be easier to talk about once we actually have some addons using them in the wild, and we can see various combinations, etc.

samselikoff:
Sounds good - thanks for cranking on all this stuff!!

PR #179, opened 2018-05-08
Use fetch in a FastBoot-compatible way

dfreeman:
This switches us to use a `docs-fetch` service rather than calling `fetch` directly, which will convert non-absolute request URLs into absolute ones taking the FastBoot environment into account. This gets us out of the way of the host app, leaving it free to use `$.ajax` or `fetch`, with or without FastBoot, however it likes.

Unfortunately we can't merge this yet as it breaks our tests. We're on a fork of ember-cli-page-object until https://github.com/san650/ember-cli-page-object/issues/397 is resolved, but even once a fix does go in, `ember-classy-page-object` has some compatibility issues with the newer API. We should also add actual tests for running in FastBoot to ensure we don't regress on that.

On a brighter note, I was able to drop Prember in on top of this change and it basically just worked out of the box, which was really nice.


samselikoff:
Nice work man 👍

dfreeman:
Note: I've updated this to use `ember-fetch-adapter`, which includes built-in Fastboot support, allowing us to drop the dependency on `ember-cli-fastboot` and get the tests passing for now.

Ultimately we should still get the page-object incompatibility sorted out and add test coverage for Fastboot, but in the interim this should at least unblock folks.

PR #180, opened 2018-05-09
Fix missing cookies on XHR GET search-index.json

ebatsi:
XHR request to fetch the search index, fails when done inside a protected domain that requires authentication (done with cookies).

### DIFF
--- a/addon/services/docs-search.js
+++ b/addon/services/docs-search.js
@@ -72,7 +72,7 @@ export default Service.extend({
 
   loadSearchIndex() {
     if (!this._searchIndex) {
-      this._searchIndex = fetch(this.get('_indexURL'))
+      this._searchIndex = fetch(this.get('_indexURL'), { credentials: 'same-origin' })
         .then(res => res.json())
         .then(json => {
           return {


ebatsi:
The build failed at the starting stage, could someone with write access restart it please?

addon/services/docs-search.js
```diff
@@ -72,7 +72,7 @@ export default Service.extend({
 
   loadSearchIndex() {
     if (!this._searchIndex) {
-      this._searchIndex = fetch(this.get('_indexURL'))
+      this._searchIndex = fetch(this.get('_indexURL'), { credentials: 'same-origin' })
```
mansona: I'm not entirely the right person to ask about these things, but my 2c is that I imagine we would want something like this to be configurable 🤔 

I assume this will make your system work and might break other people's (this is an assumption so anyone more knowledgeable please jump in!)

@ebatsi do you think you could make this optional and configurable (somehow) through the environment config? I think that might help it land a bit sooner 👍

mansona: I'm not entirely the right person to ask about these things, but my 2c is that I imagine we would want something like this to be configurable 🤔

I assume this will make your system work and might break other people's (this is an assumption so anyone more knowledgeable please jump in!)

@ebatsi do you think you could make this optional and configurable (somehow) through the environment config? I think that might help it land a bit sooner 👍

ef4:
This looks OK to me.

dfreeman:
Hi @ebatsi — sorry this has sat for so long!

We've actually been looking at incorporating [ember-fetch-adapter](https://github.com/tchak/ember-fetch-adapter) to get better out-of-the-box Fastboot support, which [sets this option by default](https://github.com/tchak/ember-fetch-adapter/blob/d642652f18276e5552304255014691483ecdfb56/addon/adapter.js#L49). I'll aim to either get that in place or land this PR sometime in the next day so that you're not stuck waiting any longer 🙂

dfreeman:
This should be fixed by #179, which I'm about to release as `v0.4.3`

ebatsi:
thanks @dfreeman, i will verify the fix using the latest version

samselikoff:
@ebatsi Closing for now. Let us know if you're still having trouble!

PR #184, opened 2018-06-01
Remove keyboard shortcuts from default template

billpull:
Something that was tripping me up typing in an input was the keyboard shortcuts were automatically included in the `application.hbs` but specified as optional.


samselikoff:
I don't think this is the default template, it's the one used for AddonDocs' own dummy app

billpull:
Oh you are right sorry didn't see that in the git history thought it was generated

PR #188, opened 2018-06-06
move latest to top-level URL

savvymas:
This PR eliminates the need to redirect from `/` to `/latest` by deploying the latest version at the top level instead of under `/latest`, moving all other versions under a special `/versions` path (like `/versions/beta` or `/versions/v1.0.0`), and adapting the redirection logic to handle both cases (top-level app vs versioned which can be distinguished by the presence of the “versions” name immediately after the true root URL).

This is a breaking change for existing sites, which leads to some questions for discussion:

- Are people OK with making this new layout default, vs having a config option that would let existing sites stay as they are?
- Should we just do a major release and give manual upgrade instructions?
- Should we try to make an automatic upgrade (it’s hard in general because deployed versioned apps have already been rewritten for their current locations, and there is no reliable pattern to search and replace).

### DIFF
--- a/lib/deploy/plugin.js
+++ b/lib/deploy/plugin.js
@@ -38,7 +38,7 @@ module.exports = class AddonDocsDeployPlugin {
   }
 
   willUpload(context) {
-    let relativeBuildDestination = this._getVersionPath();
+    let relativeBuildDestination = path.join('versions', this._getVersionPath());
     let stagingDirectory = context.addonDocs.stagingDirectory;
     let fullBuildDestination = path.join(stagingDirectory, relativeBuildDestination);
 
@@ -65,7 +65,7 @@ module.exports = class AddonDocsDeployPlugin {
       fs.copySync(`${vendorDir}/index.html`, `${stagingDirectory}/index.html`);
     }
 
-    let segmentCount = this._getRootURL().split('/').length + 1;
+    let segmentCount = this._getRootURL().split('/').length;
     let redirectContents = fs.readFileSync(`${vendorDir}/404.html`, 'utf-8');
     redirectContents = redirectContents.replace(/\bADDON_DOCS_SEGMENT_COUNT\b/g, segmentCount);
     fs.writeFileSync(`${stagingDirectory}/404.html`, redirectContents);
@@ -113,10 +113,12 @@ module.exports = class AddonDocsDeployPlugin {
   _maybeUpdateLatest(context, stagingDirectory) {
     if (!this.userConfig.shouldUpdateLatest()) { return; }
 
-    let latestDir = this._getVersionPath('latest');
+    let latestDir = '';
     let fullPath = path.join(stagingDirectory, latestDir);
-    let deployVersion = this._latestDeployVersion();
-    return fs.remove(fullPath)
+    let deployVersion = this._latestDeployVersion();    
+    return filterDir(fullPath, 
+      relativeName => ['versions', 'versions.json', '404.html'].includes(relativeName)
+    )
       .then(() => fs.copy(context.distDir, fullPath))
       .then(() => this._updateIndexContents(context, stagingDirectory, latestDir, deployVersion));
   }
@@ -133,19 +135,19 @@ module.exports = class AddonDocsDeployPlugin {
   }
 
   _currentDeployVersion() {
-    let path = this._getVersionPath();
+    let curpath = path.join('versions', this._getVersionPath());
     let name = this.userConfig.getVersionName();
     let sha = this.userConfig.repoInfo.sha;
     let tag = this.userConfig.repoInfo.tag;
-    return { path, name, sha, tag };
+    return { path: curpath, name, sha, tag };
   }
 
   _latestDeployVersion() {
-    let path = this._getVersionPath('latest');
+    let curpath = '';
     let name = 'latest';
     let sha = this.userConfig.repoInfo.sha;
     let tag = this.userConfig.repoInfo.tag;
-    return { path, name, sha, tag };
+    return { path: curpath, name, sha, tag };
   }
 
   _inferRepoUrl(context) {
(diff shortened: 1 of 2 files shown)


pzuraq:
I believe this is a duplicate of #164?

ef4:
(I helped with this PR, and think it's a good idea.)

It does look like #164 does the same thing -- what was blocking that one?

pzuraq:
I believe @dfreeman wanted to make sure that migration would be smooth, and when testing it out on one of his addons @samselikoff found that updating ec-addon-docs broke fastboot so we were unable to verify that the migration would go smoothly.

@dfreeman I think we should consider just moving forward at this point, better to get the new docs structure out. If we find that the migration is painful, we can always release patches.

ef4:
Suggestion: @savvymas and I can make the migration automatic, to help push this work forward.

I'm looking at what the other differences are between this PR and #164, and it's just the choice of `v` vs `versions` (which I don't really care about), but also #164 tries to strip leading `v` off of tag names. I get the motivation, but it's a footgun. As soon as somebody has a branch name starting with `v` they'll get a nonsensical result. 

The only other difference is that #164 updates the docs to match, which is definitely good and we should harmonize with that.

ef4:
> and when testing it out on one of his addons @samselikoff found that updating ec-addon-docs broke fastboot so we were unable to verify that the migration would go smoothly.

Are you saying that something about this change breaks fastboot, or just that there were some unrelated changes that broke fastboot that prevented testing?

dfreeman:
Sorry, I've been away from my laptop all afternoon.

> #164 tries to strip leading v off of tag names. I get the motivation, but it's a footgun. As soon as somebody has a branch name starting with v they'll get a nonsensical result.

To be clear, it only strips the `v` off [if it's immediately followed by a number](https://github.com/ember-learn/ember-cli-addon-docs/pull/162/files#diff-f7bf2b665273dfa66ffa4ad7c0a52bb9R26). That code is also in the default implementation of a user-overridable hook, so it's simple to change in cases where people do want other behavior.

I'm definitely not married to the idea, but it does feel nicer personally. It's not obvious to me that there are cases where a tag or branch name starting with `v<digit>` would lead to a weird path after stripping the `v`, but it's entirely possible I'm missing something 🙂 

>  I think @dfreeman's PRs have logic for the conversion process that he thought would work, so it may just be that we need to rebase them and verify that everything works as expected.

Neither of my PRs have any code for doing the migration—I only documented the rough set of steps I'd followed myself when moving test sites over.

> Suggestion: @savvymas and I can make the migration automatic, to help push this work forward.

@ef4 @savvymas if you're up for that, that would be fantastic!

In my mind the primary blocker to this point was that the only person who'd successfully migrated a site over was also both the author of the only current docs for doing so _and_ the implementer of the code in question. I wanted to make sure that I hadn't left something out or glossed over something confusing with the actual implementation fresh in my mind.

samselikoff:
I’ve since fixed the unrelated fastboot issue so I can give it a shot.

ef4:
We have the auto migrator code nearly ready, it would be great to get more testing for that too. 

One thing it does that I think is important is installing permanent redirection support for all the URLs that break after migration. Breaking URLs is bad, which is why I’ve chosen to treat this issue as a blocker for publishing the ember-animated docs.

samselikoff:
@ef4 ping me when the migrator is ready and I can try it out on Storefront!

samselikoff:
I'd like to see `v` instead of `versions`, is that possible here?

(Because that determines the URLs correct? Something like `ember-cli-mirage.com/v/4.2.0`?)

samselikoff:
Agreed - also, I don't think the migrator should block us, so if we decide I can test it out manually on Storefront.

We can hold off releasing until there's a migrator, to not break URLs.

ef4:
I think the thing we need to agree on is the URL structure. As for the PRs, I will merge them together and review to make sure we have the best of both.

I don't really care about `v` vs `versions`. I am still slightly negative on having a rule for mangling tag names before putting them into URLs, but if everybody else really likes it I can live with it.

dfreeman:
If we use `v` as our marker segment, then I'm in favor of the tag mangling — I'm not a fan of `v/v1.2.3`. Beyond that, I don't have strong opinions.

samselikoff:
Agree with Dan I think:

```
ember-cli-mirage.com/v/0.4.1/

ember-cli-mirage.com/v/v0.4.1/

ember-cli-mirage.com/versions/0.4.1/

ember-cli-mirage.com/versions/v0.4.1/
```

1 is my favorite.

dfreeman:
For completeness, we should consider any scheme we look at with branch names too, e.g.

```
ember-cli-mirage.com/v/master/

ember-cli-mirage.com/versions/master/
```

samselikoff:
Right... since we're on the topic, did we ever discuss having a `/branches` namespace?z

ef4:
There's nothing about addon-docs itself that is tied to branches or tags. It's all just "builds", and it seems better to me not to make assumptions about the structure of builds.

```
http://ember-cli-mirage.com/installation
http://ember-cli-mirage.com/builds/v1.0.0/installation
http://ember-cli-mirage.com/builds/experimental/installation
http://ember-cli-mirage.com/builds/oldstable/installation
http://ember-cli-mirage.com/builds/latest-nightly/installation
```

Just to take one example: you may end up supporting multiple channels (oldstable and stable). You 
may want the `oldstable` URL to always point at the newest release on that channel (not the branch).

There's just a lot of ways to do it.

samselikoff:
Yep I agree having the point of abstraction is nice. (I think the term `branches` could be more clear for things like PR builds but I still think it's worth the tradeoff.)

```
http://ember-cli-mirage.com/versions/v1.0.0/installation
http://ember-cli-mirage.com/versions/latest/installation

http://ember-cli-mirage.com/builds/v1.0.0/installation
http://ember-cli-mirage.com/builds/latest/installation
```

Now I'm leaning towards `versions` or `builds` haha. What do you all think?

dfreeman:
100% agreed that trying to impose a branches/tags structure on things gets in the way of real-world usage patterns. I have some hesitation around the world `build`, though — that seems to be framing things in terms of 'builds of the documentation' rather than 'versions of the addon'.

samselikoff:
Let's go with `versions`

PR #190, opened 2018-06-10
Eliminate free-floating 'latest' and 'master' references

### LINKED ISSUE #181: Allow ability to configure branch name
Currently, the [`{{x-main}}`](https://github.com/ember-learn/ember-cli-addon-docs/blob/master/addon/components/docs-viewer/x-main/component.js#L79-L86) component in the `{{docs-viewer}}` has hard-coded URLs that assume that the main branch of the repo is always `master`. I use `latest` for my repositories, and would like the "Edit docs" link to work 😅 

Any guidance as to the best way to achieve this is welcome, as I'd be up to making a PR for this particular issue.

dfreeman:
Building on the work to move the deploy of the latest release into the root directory, this PR eliminates places where we have hard-coded specialized behavior for versions called `latest` or `master`. For the former, I've changed the key to `-latest` and moved it into a shared constant. For the latter, `master` is now the default value for a `getPrimaryBranch()` hook that devs can override.

Using `-latest` rather than `latest` should help to some degree in mitigating the likelihood of name clashes with actual branches, and since we're due for a breaking release, this seems like the time to make that change. Allowing for configuration of the primary branch should fix #181 and help folks using strategies like [git-flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) where the primary branch is `develop` rather than `master`.

@ef4 @savvymas where do things stand with the auto-migration code? Ideally we'd incorporate the `latest` -> `-latest` change in `versions.json` there—I'm happy to add that whenever the script is ready to share.

### DIFF
--- a/addon/components/docs-header/component.js
+++ b/addon/components/docs-header/component.js
@@ -7,7 +7,7 @@ import { addonLogo } from 'ember-cli-addon-docs/utils/computed';
 import { inject as service } from '@ember/service';
 import { reads } from '@ember/object/computed';
 
-const { projectName, projectHref } = config['ember-cli-addon-docs'];
+const { projectName, projectHref, latestVersionName } = config['ember-cli-addon-docs'];
 
 /**
   Render a header showing a link to your documentation, your project logo, a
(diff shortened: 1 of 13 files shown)


ef4:
Yes, we can do that. Should have a PR for you in a day or two.

I'll make sure it merges this PR and #164 so it is comprehensive.

dfreeman:
Superseded by #196

PR #191, opened 2018-06-11
Fix bug where the token separators used in the search component and the search indexer were inconsistent

antidis:
This meant that words like `docs-snippet` were being indexed as `docs` and `snippet`, but would be searched for as `docs-snippet`.

I'm not sure that method of sharing state is particularly great, but we can't `import` from `lib/broccoli`, or `require` from `addon/services`.

There's some good suggestions over here on how search could be further improved with a Lunr pipeline: https://github.com/olivernn/lunr.js/issues/296

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,7 +47,8 @@ module.exports = {
         projectName: this.parent.pkg.name,
         projectTag: this.parent.pkg.version,
         projectHref: info && info.browse(),
-        deployVersion: 'ADDON_DOCS_DEPLOY_VERSION'
+        deployVersion: 'ADDON_DOCS_DEPLOY_VERSION',
+        searchTokenSeparator: /\s+/
       }
     };
 
--- a/index.js
+++ b/index.js
@@ -48,7 +48,7 @@ module.exports = {
         projectTag: this.parent.pkg.version,
         projectHref: info && info.browse(),
         deployVersion: 'ADDON_DOCS_DEPLOY_VERSION',
-        searchTokenSeparator: /\s+/
+        searchTokenSeparator: "\\s+"
       }
     };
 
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -47,7 +47,8 @@ module.exports = {
         projectName: this.parent.pkg.name,
         projectTag: this.parent.pkg.version,
         projectHref: info && info.browse(),
-        deployVersion: 'ADDON_DOCS_DEPLOY_VERSION'
+        deployVersion: 'ADDON_DOCS_DEPLOY_VERSION',
+        searchTokenSeparator: /\s+/
```
dfreeman: Unless I'm mistaken, a regex won't survive the `JSON.stringify` pass that happens when config is serialized for the app. Can we add a test to verify this is working?

antidis: Works fine for the dummy app:

screen shot 2018-06-11 at 15 35 04

I'll get a test in place.

antidis: Added. I'm open to a different location for the test.

tests/acceptance/sandbox/api/components-test.js
```diff
@@ -45,4 +45,12 @@ module('Acceptance | API | components', function(hooks) {
     assert.equal(indexItems.length, 13, 'correct number of items rendered');
     assert.ok(indexItems.includes('_privateField'), 'private field rendered');
   });
+
+  test('search box works', async function(assert) {
+    await visit('/sandbox');
+
+    assert.equal(modulePage.searchResults.items.length, 0, 'no search results shown');
+    await modulePage.fillInSearchQuery('sub-subsection');
```
antidis: Oddly enough, the search works as every character of `sub-subsection` is typed out, except when the query is `sub-subsecti` or `sub-subsectio`.

index.js
dfreeman: It seems like that's working because the search terms aren't being split at all — that means if I search e.g. `render multiple` with your changes, I get no results, but on the latest deployed version of the site, that returns several options back.

It looks like the regex is being encoded as `{}` in the serialized config:

screen shot 2018-06-11 at 1 00 46 pm

antidis:
Pushed a change to reencode the RegExp as a string, that then gets converted to a RegExp.

PR #192, opened 2018-06-12
Prevent keyboard shortcuts from running when the document's active element is a form element

antidis:
When using some of the old examples on the demo page, I found that the page would change location if I hit J or K. This was quite annoying, so this is an attempt to fix that problem.

Although live examples have been removed, I think this is worth adding. Snippets may have demos with form inputs, and it would be messed up to have the page changed unexpectedly when typing.

### DIFF
--- a/addon/components/docs-keyboard-shortcuts/component.js
+++ b/addon/components/docs-keyboard-shortcuts/component.js
@@ -4,6 +4,7 @@ import { later } from "@ember/runloop";
 import layout from './template';
 import { EKMixin, keyUp } from 'ember-keyboard';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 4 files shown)


pzuraq:
I think it would be cleaner to have a checker function that returns true/false rather than calling a callback:

```js
    if(formHasFocus()) {
      this.set('isGoingTo', true);
      later(() => {
        this.set('isGoingTo', false);
      }, 500);
    }
```

But I'm onboard with the behavior change.

antidis:
Updated to use a function instead.

antidis:
The callback method is something I've used a lot, mostly in cases where you may not have the object you're trying to run the function on though.

dfreeman:
Thanks @antidis!

PR #194, opened 2018-06-13
[bugfix] Prevents all Marked escaping in plain text

pzuraq:
This ensures that _all_ escaped characters are unescaped when rendering
markdown. Should minimize the impact Marked has on our templates. I've
also opened an issue with Marked directly to see if we can upstream
this.


dfreeman:
Can we include test coverage for the cases this is fixing?

PR #196, opened 2018-06-16
Site reorg including top-level latest

ef4:
This rolls up and reconciles work in #190, #188, #164, #163, and #162.

It includes auto-migration support that updates existing sites to the new format. We tested it using ember-cli-addon-docs own site as the test case.


ef4:
I think this is ready to go. The last bit that held things up was making the migration support the `latest` -> `-latest` PR. The problem is that versions.json as implemented is not very upgradeable because too much of the semantics are implemented in the components as opposed to the build.

I changed the implementation of that PR to move the logic for identifying the latest out of the client and into the build, which prevents old versions from showing a literal `-latest` in the version selector. But there was already some sorting logic in the version selector component in those earlier versions which we can't readily patch over, so they will end up with `Latest` at the bottom instead of the top.

It's possible to redesign versions.json as a more upgradeable format. Ideally the component would be entirely logicless and all sorting and choice of display titles would happen during the build and get encoded as versions.json. But that would be even more disruptive and I really think we should try to land all this in-flight work.

package.json
```diff
@@ -31,6 +31,7 @@
     "broccoli-plugin": "^1.3.0",
     "broccoli-source": "^1.1.0",
     "broccoli-stew": "^1.5.0",
+    "ember-ace": "^1.2.0",
```
dfreeman: I don't think this should have been added back in

index.js
```diff
@@ -47,6 +52,8 @@ module.exports = {
         projectName: this.parent.pkg.name,
         projectTag: this.parent.pkg.version,
         projectHref: info && info.browse(),
+        primaryBranch: userConfig.getPrimaryBranch(),
+        latestVersionName: LATEST_VERSION_NAME,
```
dfreeman: Should this be `latestVersionKey`/`LATEST_VERSION_KEY` now?

addon/services/project-version.js
```diff
@@ -2,17 +2,21 @@ import Service, { inject as service } from '@ember/service';
...
 
   _loadAvailableVersions: task(function*() {
     let response = yield this.get('docsFetch').fetch({ url: `${this.get('root')}versions.json` }).response();
-    let json = yield response.ok ? response.json() : { latest: this.get('currentVersion') };
+    let json = yield response.ok ? response.json() : { [latestVersionName]: this.get('currentVersion') };
 
     this.set('versions', Object.keys(json).map(key => {
       let version = json[key];
       version.truncatedSha = version.sha.substr(0,5);
+      version.key = key;
```
dfreeman: Between this and all the `delete version.key` stuff going on in the deploy plugin, it seems like `key` is coming and going from these objects a lot. Distinguishing between `key` and `name` seems like a Clearly Good Thing to me, but it feels like there's a lot of data munging going on to accomplish that, and I don't think I understand the flow well enough to see why it's necessary.

index.js
ef4: Yes, probably worth renaming.

addon/services/project-version.js
ef4: IMO the problem is that the `versions.json` format is really not right. It should be a list, not a POJO. Much code here (both this munging that I added and several existing mungings) is working around that fact. The Ember service was already exposing the versions as a list anyway, since that's what it clearly wants to be. This setting and deleting of `key` is how we translate between POJO:

`{ "-latest": { sha: "xxx", name: "Latest } }`

and list:

`[{ sha: "xxx", name: "Latest", key: "-latest" }]`

I took it as a constraint that I shouldn't break earlier builds that use earlier versions of addon-docs, which meant sticking with the format we have.

The key vs name distinction is needed because we want to label some version with the latestVersionName (or latestVersionKey), but if we overload `name` for that purpose, we require logic in the ember app to replace that name with `"Latest"`, but logic in the ember app _is not upgradeable_, in the sense that older existing builds will keep doing whatever they're already doing. They were literally printing `"-latest"`.

I would be happy to switch to a list format for versions.json (and make the version-selector component even dumber, particularly by taking away its responsibility to sort), but there's no way to do that without breaking existing sites, short of maintaining both versions.json and versions2.json for a long time. Maybe that is what we have to do.

We're dealing with a stronger form of breaking changes here than normal. It's not enough to just do a major release of addon-docs and break a bunch of things. To get over a versions format change, people would need to go back to every tagged build they still care about documenting, upgrade that dummy app to latest addon-docs (probably non-trivial work), and manually re-deploy it.

dfreeman: > The key vs name distinction is needed because we want to label some version with the latestVersionName (or latestVersionKey), but if we overload name for that purpose, we require logic in the ember app to replace that name with "Latest", but logic in the ember app is not upgradeable, in the sense that older existing builds will keep doing whatever they're already doing. They were literally printing "-latest".

Yep, I get the problem and am on board with adding the distinction as a way to fix it 👍 

What threw me was creating the hash with the `key` key, immediately removing that key, and then adding it back in at runtime. An additional key on the version objects shouldn't break old deploys, so that we could just leave that in (and backfill it for older versions as part of the migration script) and simplify both the deploy and the runtime code a bit, right?

> We're dealing with a stronger form of breaking changes here than normal. It's not enough to just do a major release of addon-docs and break a bunch of things. To get over a versions format change, people would need to go back to every tagged build they still care about documenting, upgrade that dummy app to latest addon-docs (probably non-trivial work), and manually re-deploy it.

For sure. We've already brushed up against this a few times, and the severity of a breaking change like this only gets worse as time goes on. I think I feel much less strongly than you that the current `versions.json` format is problematic (I get that making the versions selector as dumb as possible would mitigate the issue we're facing here, but I'm not totally sold that the order of a machine-generated list should be the source of truth for the human-consumable set of versions), but I certainly wouldn't oppose the change if you or anyone wanted to champion it.

ef4: > but I'm not totally sold that the order of a machine-generated list should be the source of truth for the human-consumable set of versions

That's not what I'm suggesting. I'm suggesting we take the [existing sorting logic](https://github.com/ember-learn/ember-cli-addon-docs/blob/b4bfe91ad58406808b960fc1870d571b16b462a4/addon/components/docs-header/version-selector/component.js#L15-L21) and move it out of the ember app and into the build code that writes the versions list. This would ensure that the version-selector menu doesn't reorder itself as you move from version to version.

dfreeman: 👍 I see — definitely not opposed to the idea, but agreed that it's a hairy proposition given where we are currently.

pzuraq: To be fair, we are still pre-1.0. I don't think we have many users, or many versions of sites deployed. Definitely would like to make sure whatever we land on next is ideal and we won't have to update it again, but I wouldn't have any issues going back and updating the previous versions of my docs sites - that's the price of early adoption 😄 

I think using an array is a solid approach here, seems like it's pretty future proof. We can control the logic in the build, and any future changes we'd like to make can be handled there.

Issue #204, opened 2018-07-15
New component that renders API docs for a project


gossi:
For now `{{docs-viewer.nav}}` helds the API reference. I propose to decouple this into a separate component:

```hbs
{{! /addon/components/docs-viewer/x-api}}

{{!-- Autogenerated API docs --}}
      {{#if (gt project.modules.length 0)}}
        {{docs-viewer/x-section 'API Reference' style='large'}}

        {{#each-in project.navigationIndex as |type items|}}
          {{docs-viewer/x-section type}}

          {{#each items as |item|}}
            {{docs-viewer/x-nav-item (break-on item.name '/') (concat root '.api.item') item.path}}
          {{/each}}
        {{/each-in}}
    {{/if}}
```

Which will be invoked as `{{docs-viewer.api @project=myProject @root='myproject'}}`. It wouldn't even change the way the current component works, it just invokes the new component.

This would change the flexibility in which we can create docs. Some which I would need at the moment, because I run into this ;)

RobbieTheWagner:
I'm closing some old issues. If you think this is still valid, please feel free to open an updated issue and we can try to prioritize it.

PR #205, opened 2018-07-18
Add {{docs-viewer/x-api}} component

gossi:
Implementation for #204


pzuraq:
I'm not sure this is something we want to make public API just yet, especially given API docs would be added anyways. Users using the contextual component would end up having two API doc nav menus, which is not ideal.

We've had some discussions about how to make API docs more flexible in general. One idea was to move the API docs section to a top level tab, so that it's not in the left navigation bar at all. I think this would be better in quite a few cases, especially for addons with large surface areas, but we're still not sure.

I'd also like to extract the API docs into an engine, I started doing it one day and got about halfway done but ran out of time to work on it.

gossi:
I made it optional to use the api component (the current behavior is untouched). At the moment I write docs for an addon suite. "Big" parts go individually into their own project (those links in the header). Smaller parts will be grouped under one link in the header. Following this one, will have a block for each addon on the left. Each block with an API reference. That's the reason for this PR.

pzuraq:
Yeah, I think we should go back to the drawing board and try to solve this in a standardized way. We do want to support larger projects, but we have plans to do this in a different way (via a meta site that allows us to search docs for many addons, all built independently, like https://docs.rs/ or https://www.rubydoc.info/). 

If that UX is not good enough, I think we need to come up with a better solution, probably at the build level. The user should not have to implement the navigation themselves.

gossi:
Yeah esdoc also jumped into this: https://doc.esdoc.org/

For the time being I will write a component for my docs app 👍

pzuraq:
Closing this for now

PR #207, opened 2018-07-24
Build order agnostic

### LINKED ISSUE #200: Template processing order issue with Ember CLI v3.2
Getting this error when upgrading to Ember CLI 3.2: `Templates contents haven't been extracted yet`.

It might be that ec-addon-docs was relying on the order that Ember CLI was processing the templates.

Here is the full log: https://gist.github.com/josemarluedke/e86327831aeae56fd62b17309a7ed56f

dfreeman:
This should fix #200 — it's a hack, but it's at least an encapsulated one, and the ordering assumptions we were making before were no less hacky, just more implicit.


josemarluedke:
Just tested out this branch in my project and it definitely fixes the problem. Thanks for working on this.

dfreeman:
👍 we need to get our flakey test situation figured out...

PR #208, opened 2018-07-24
Allow child routes under docs pages

savvymas:
(authored while pairing with @ef4)

This makes the docs-route service more tolerant of pages with their own internal URL structure. For example, we're using this in ember-animated to have a page that includes an outlet in order to demonstrate cross-route animations.

### DIFF
--- a/tests/dummy/app/router.js
+++ b/tests/dummy/app/router.js
@@ -12,7 +12,11 @@ Router.map(function() {
     this.route('usage');
     this.route('quickstart');
     this.route('patterns');
-    this.route('deploying');
+    this.route('deploying', function() {
+      // This exists so we can acceptance test whether the docs pages tolerate
+      // embedded child routes. It's not used in the public documentation site.
+      this.route('test-nested-route');
+    });
 
     this.route('components', function() {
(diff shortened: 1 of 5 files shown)


tests/dummy/app/router.js
```diff
@@ -12,7 +12,11 @@ Router.map(function() {
     this.route('usage');
     this.route('quickstart');
     this.route('patterns');
-    this.route('deploying');
+    this.route('deploying', function() {
+      // This exists so we can acceptance test whether the docs pages tolerate
+      // embedded child routes. It's not used in the public documentation site.
+      this.route('test-nested-route');
```
pzuraq: Can we add this to the sandbox app instead? The sandbox app is meant to be our "scenarios" app, where we put things that are dummy content which only exist for testing and demonstration purposes

ef4: Oh yeah, definitely, I didn't know that existed and felt vaguely bad about this already.

ef4:
I moved the test route under sandbox, so this should be ready now.

samselikoff:
Sweet! Thanks for all the contribs ^_^

PR #209, opened 2018-07-24
Upgrade ember-svg-jar

savvymas:
This version has many bugfixes, and we're already using it in ember-animated docs, so this eliminates a conflict for us.

There were no breaking changes.

(authored while pairing with @ef4)


dfreeman:
Thanks @savvymas!

Issue #178, opened 2018-05-07
Dealing with css conflicts


Dhaulagiri:
The addon I am writing docs for uses a design system that is based on Tachyons.  This is causing me problems in a few cases, most notably with the `absolute` class added to the nav [here](https://github.com/ember-learn/ember-cli-addon-docs/blob/4136850772b9aafd3b381014b2a5b36e67b9e6c8/addon/components/docs-viewer/x-nav/template.hbs#L12).  The problem is that because I'm importing _my_ css first before the css from this addon _my_ `absolute` is superseding positioning applied to the nav by the `lg:relative` class and therefore always positioning the nav absolutely. 

I _think_ there might be a way to fix this particular case, but more broadly do you have any advice on how to address this kind of issue?

samselikoff:
Yes - we should probably prefix the Tailwind classes we're using, perhaps with `ad-` so they don't conflict.

I also want to get PurgeCSS in here so we don't ship a bunch of unused classes.

acdn-sglanzer:
+1 on this - I'm also using a modified Tachyons system and I've been overriding templates to "fix" the class conflicts

MelSumner:
Looks like some stuff is styled on the html elements which can also cause issues: 
![image](https://user-images.githubusercontent.com/4587451/44002713-d2829274-9e0c-11e8-9106-6eb483f66ad3.png)

olivier-w:
It looks like tailwind also adds a `border-color` to [every single element](https://github.com/tailwindcss/tailwindcss/blob/master/css/preflight.css#L535).

Is it possible to completely exclude tailwind from ember-cli-docs?
```
*, *::before, *::after {
    border-width: 0;
    border-style: solid;
    border-color: #dae1e7;
}
```

pzuraq:
We do want to provide a consistent theme that addons can use conventionally. The long term goal is to create a unified ecosystem of documentation, so Ember devs can browse docs quickly and effectively (possibly through a single unified website).

If Tailwind has too much overlap with user’s styles, that is tricky, but we can try to change it. I think giving options to disable them encourages users to write their own styles instead, which is counter to that long term goal.

samselikoff:
@olivier-w this one is from `normalize.css` which we can definitely add an option to exclude. Curious in this case what problem you're running into with this rule? Is it because you're adding a `border-width` but then having to change the color?

samselikoff:
[...]

PR #212, opened 2018-07-26
Use ad- tailwind prefix

jrowlingson:
Addresses https://github.com/ember-learn/ember-cli-addon-docs/issues/178

> Yes - we should probably prefix the Tailwind classes we're using, perhaps with ad- so they don't conflict.

cc @samselikoff

### DIFF
--- a/addon/components/api/x-import-path/component.js
+++ b/addon/components/api/x-import-path/component.js
@@ -4,7 +4,7 @@ import hbs from 'htmlbars-inline-precompile';
 export default Component.extend({
   tagName: '',
   layout: hbs`
-    <div class="import-path mb-6 font-mono" data-test-import-path>
+    <div class="import-path ad-mb-6 ad-font-mono" data-test-import-path>
       <span class="hljs-keyword">import</span>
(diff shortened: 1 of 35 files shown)


samselikoff:
@pzuraq @dfreeman I'm 👍 on this but understand why people might hate me for it 😄 

I have a future in mind where this sort of thing is done at build-time, so we still get to use vanilla classes but they're prefixed when the addon is used by others. But for now I think this change makes us a better citizen.

In retrospect Tailwind may not have been the best choice for styling this addon for these sorts of issues (it's hard to contain, based on global utilities, etc.) but I still do think it has many strengths for a project like this, notably preventing css debt from growing over time.

Hopefully I (or someone else) can eventually add these build-time features to ember-cli-tailwind in the future.

pzuraq:
I 👍 as well, the only thing that bothers me occasionally about Tailwind is it makes customizing addon-docs' CSS hard, but tbh that's a good thing because it makes me fix things upstream more often 😛

Let's move forward with this for now

PR #213, opened 2018-08-02
classes on doc-demo component

miguelcobain:
This should allow us to customize the classes added to the docs-demo component.


pzuraq:
👍 I'm ok with this, @samselikoff thoughts?

samselikoff:
lgtm!

PR #214, opened 2018-08-08
Delete .DS_Store

miguelcobain:


ro0gr:
Maybe add it to the gitignore as well?

PR #220, opened 2018-08-16
Upgrade ember-code-snippet

dcyriller:
To be compatible with Broccoli 1 & 2

This PR would remove some of these deprecation warnings:
```
[API] Warning: The .read and .rebuild APIs will stop working in the next Broccoli version
[API] Warning: Use broccoli-plugin instead: https://github.com/broccolijs/broccoli-plugin
[API] Warning: Plugin uses .read/.rebuild API: 
[API] Warning: Plugin uses .read/.rebuild API: 
```


pzuraq:
Thanks for the patch!

PR #231, opened 2018-08-26
Make generated API reference its own top-level section

xcambar:
Following up comments on https://github.com/ember-learn/ember-cli-addon-docs/pull/230
I gave a shot at working on moving API docs to their own top-level section.

This PR provides the following significant features:

* zero-config API tab for the projects that have auto-generated API docs
* A new `API Reference` section pointing to... well... the API reference
* This link is **not** displayed when the project does not have API docs (useful for smaller addons)
* The API reference is under `/api` instead of `/docs/api`
* The navbar under `/api` and `docs` differ in their contents (I assumed it was expected, but IMO it gives much more readable structure)
* The section of the navbar pointing to the API reference has disappeared from the navbar in `/docs`

:warning This PR does NOT provide tests, they'll be added if/once the feature set is agreed upon.

@samselikoff @pzuraq this should match your comments. Hope you'll like it. If so, this will close #230.


xcambar:
Here's the diff to be applied to `ember-cli-mirage` (see https://github.com/samselikoff/ember-cli-mirage/pull/1377) to benefit from this PR:

```diff
diff --git a/tests/dummy/app/router.js b/tests/dummy/app/router.js
index 01b4a52..fc2c8c7 100644
--- a/tests/dummy/app/router.js
+++ b/tests/dummy/app/router.js
@@ -25,10 +25,6 @@ Router.map(function() {
       this.route('1-belongs-to');
       this.route('2-has-many');
     });
-
-    this.route('api', function() {
-      this.route('class', { path: '/:class_id' });
-    });
   });
 
   // this.route('contacts', { path: '/' });
```

So it's actually less code to be written by the developer 😄

PR #237, opened 2018-09-04
Update 'Edit this page' link regex

22a:
This PR updates the `Edit this link` filter regex from
* `template.+(hbs|md)`
to
* `\.(hbs|md)$`

this removes the necessity to have the string "template" in doc filenames, in turn preventing dead links such as https://github.com/user/repo/edit/branch/undefined

I'm not too concerned about the trialing `$` if it's not agreeable but happy to discuss.

### DIFF
--- a/addon/components/docs-viewer/x-main/component.js
+++ b/addon/components/docs-viewer/x-main/component.js
@@ -80,7 +80,7 @@ export default Component.extend({
       }
     } else {
       let file = appFiles
-        .filter(file => file.match(/template.+(hbs|md)/))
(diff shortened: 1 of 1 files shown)


dfreeman:
Thanks @22a! This seems reasonable to me, but @samselikoff should probably confirm. I wonder if there's a lightweight way to include test coverage for this so we don't regress in the future?

samselikoff:
Even a hard-coded test for a currently existing docs page would be great

22a:
@dfreeman @samselikoff Good point, added!

PR #240, opened 2018-09-10
Omit <a> title attribute if title is null

22a:
This PR tweaks the `title()` method of the `HBSRenderer` in [lib/utils/compile-markdown.js](https://github.com/22a/ember-cli-addon-docs/blob/b75247734aff43f30a63f51e090f8d60a857ba02/lib/utils/compile-markdown.js#L124-L127) to only add a title attribute to the html snippet when the passed `title` argument is truthy. 

This prevents adding `null` (and perhaps also `undefined`, but I haven't encountered that yet) title attributes to `a` tags, such as this ([visible here](https://ember-learn.github.io/ember-cli-addon-docs/docs)):

image of the string 'null' title attribute visible when hovering over a link

---

This PR also bumps the [`marked`](https://github.com/markedjs/marked) dependency from `^0.3.12` to `^0.5.0` for no particular reason other than to keep it up to date - I don't feel strongly about this and I'm happy to remove this change from the PR if you'd prefer.

### DIFF
--- a/lib/utils/compile-markdown.js
+++ b/lib/utils/compile-markdown.js
@@ -122,6 +122,7 @@ class HBSRenderer extends marked.Renderer {
   }
 
   link(href, title, text) {
-    return `<a href="${href}" title="${title}" class="docs-md__a">${text}</a>`;
+    const titleAttribute = title ? `title="${title}"` : '';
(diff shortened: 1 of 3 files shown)


22a:
cc @samselikoff @dfreeman

samselikoff:
Thank you!

Issue #242, opened 2018-09-10
Ember data dependency


ryanto:
I have an issue I'm running into that I'm not quite sure how to solve. I'll try my best to explain, it's a little fuzzy exactly what's going on.

I'm writing an addon that is designed to work alongside Ember Data. On travis, my Ember try config lists multiple versions of Ember and Data. For example...

```js
// config/ember-try.js

// ...
{
  name: 'ember-lts-2.12',
  npm: {
    devDependencies: {
      'ember-source': '~2.12.0',
      'ember-data': "~2.12.0"
    }
  }
},
// ...
{
  name: 'ember-3.4',
  npm: {
    devDependencies: {
      'ember-source': '~3.4.0',
      'ember-data': "~3.4.0"
    }
  }
},
// ...
```

However, the versions of data listed above are not the versions of data that end up in my apps dummy build. It looks like the ED that ultimately ends up in the build is the version that comes from addon docs.

For example, when I run the ember-lts-2.12 scenario, I end up with an application that has addon docs 0.5.3 and Ember Data 2.18.4.

When I run yarn list I see...

```text
yarn list ember-data
├─ ember-cli-addon-docs@0.5.3
│  └─ ember-data@2.18.4
└─ ember-data@2.12.2
```

And when I run `DS.Version` in the browser I see

```
DS.VERSION
=> "2.18.4"
```

Next, when I run the ember-3.4 scenario, I end up with an application that also has addon docs 0.5.3 and Ember Data 2.18.4.

```text
yarn list ember-data
├─ ember-cli-addon-docs@0.5.3                                                                                       │
│  └─ ember-data@2.18.4                                                                                             │
└─ ember-data@3.4.0
```

And in the browser...

```
DS.VERSION
=> "2.18.4"
```

I'm wondering if you would be open to loosening the ember data dependency? Would `2.x - 3.x` be ok?

samselikoff:
#243 closes

PR #243, opened 2018-09-11
Ember data 2 or 3

ryanto:
This is a PR for Issue https://github.com/ember-learn/ember-cli-addon-docs/issues/242

It allows addon docs to accept a wider range of Ember Data versions. This way it won't conflict with the host addon's data dependency.


samselikoff:
Thank you my dude 👊

PR #248, opened 2018-09-27
Add prompt for ESDoc/YUIDoc

rwwagner90:
This adds a prompt so you can select ESDoc or YUIDoc when the blueprints run.

### DIFF
--- a/blueprints/ember-cli-addon-docs/index.js
+++ b/blueprints/ember-cli-addon-docs/index.js
@@ -72,9 +72,6 @@ module.exports = {
             }
           });
         });
-      return this.addAddonsToProject({
-        packages: ['ember-cli-addon-docs-yuidoc']
-      });
     }
   },
 
--- a/blueprints/ember-cli-addon-docs/index.js
+++ b/blueprints/ember-cli-addon-docs/index.js
@@ -22,7 +22,7 @@ module.exports = {
     });
   },
 
-  afterInstall() {
+  afterInstall(options) {
     let configPath = require.resolve(this.project.configPath());
     let configContents = fs.readFileSync(configPath, 'utf-8');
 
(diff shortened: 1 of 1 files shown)


blueprints/ember-cli-addon-docs/index.js
```diff
@@ -63,9 +63,39 @@ module.exports = {
     }
 
     if (!hasPlugins) {
+      return this._chooseAddonsToInstall()
```
dfreeman: Looks like there's a lint failure because the `return` below this one is dead code now

blueprints/ember-cli-addon-docs/index.js
```diff
@@ -63,9 +63,39 @@ module.exports = {
...
+        .then((addon) => {
+          return this.addAddonsToProject({
+            packages: [addon],
+            blueprintOptions: {
+              save: options.save
```
dfreeman: What is this option doing? Was something behaving wrong when we didn't have it before?

blueprints/ember-cli-addon-docs/index.js
```diff
@@ -63,9 +63,39 @@ module.exports = {
...
       });
     }
+  },
+
+  _chooseAddonsToInstall() {
+    // Ask which ember addon to install
+    return this.ui.prompt({
+      type: 'list',
+      name: 'addonToInstall',
+      message: 'Which documentation style would you like to use?',
```
dfreeman: Should we include a choice not to install anything at all? I could imagine wanting to use another addon-docs plugin that isn't necessarily part of the ember-learn org, or just not needing or wanting generated API docs for an addon at all.

dfreeman: Fair enough. I'll probably turn around and add a "no plugin" option anyway because I think it's valuable, but I don't need to push that work on you here 🙂

PR #258, opened 2018-10-07
Inform to github that scripts/parallel is vendored

josemarluedke:
This apparently worked for ember-auto-import: https://github.com/ef4/ember-auto-import/blob/master/.gitattributes


dfreeman:
Thanks @josemarluedke!

PR #266, opened 2018-10-16
Fix issues with empty rootURL

### LINKED ISSUE #265: Issues with empty rootURL
When using a custom domain for a project app, i'm overriding the `getRootURL()` function as stated in the [docs](https://ember-learn.github.io/ember-cli-addon-docs/docs/deploying#getrooturl-)

However, this create incorrect assets path in the index.html because of this line: 
https://github.com/ember-learn/ember-cli-addon-docs/blob/731830a208e6a36562a4639867439649720a032a/lib/deploy/plugin.js#L156

With `let rootURL = [this._getRootURL(), appRoot].filter(Boolean).join('/');` set to an empty string, the asset path become `//assets...` which is considered as a domain.

Is there a quick work around for this ?

tempo22:
This will close #265 

Incorrect segmentCount was calculated and doubles slashes were inserted when using an empty string as rootURL (as the doc suggest to do with a custom domain configuration)

### DIFF
--- a/lib/deploy/plugin.js
+++ b/lib/deploy/plugin.js
@@ -64,7 +64,7 @@ module.exports = class AddonDocsDeployPlugin {
 
   _verifyRedirectFile(stagingDirectory) {
     let vendorDir = `${__dirname}/../../vendor/ember-cli-addon-docs`;
-    let segmentCount = this._getRootURL().split('/').length;
+    let segmentCount = this._getRootURL().split('/').filter(Boolean).length;
     let redirectContents = fs.readFileSync(`${vendorDir}/404.html`, 'utf-8');
     let redirects = this._discoverOldPaths(stagingDirectory);
 
--- a/lib/deploy/plugin.js
+++ b/lib/deploy/plugin.js
@@ -150,10 +150,10 @@ module.exports = class AddonDocsDeployPlugin {
 
   _updateIndexContents(context, stagingDirectory, appRoot, deployVersion) {
     let indexPath = `${stagingDirectory}/${appRoot}/index.html`;
-    let rootURL = [this._getRootURL(), appRoot].filter(Boolean).join('/');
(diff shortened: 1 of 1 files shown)


lib/deploy/plugin.js
```diff
@@ -150,10 +150,10 @@ module.exports = class AddonDocsDeployPlugin {
 
   _updateIndexContents(context, stagingDirectory, appRoot, deployVersion) {
     let indexPath = `${stagingDirectory}/${appRoot}/index.html`;
-    let rootURL = [this._getRootURL(), appRoot].filter(Boolean).join('/');
+    let rootURL = [this._getRootURL(), appRoot].filter(Boolean).join('/').replace(/^\/?([^/]+(?:\/[^/]+)*)\/?$/, '/$1/') || '/';
```
pzuraq: Think you could add a comment explaining this regex? It's pretty complicated, having a hard time trying to parse it and understand what it's doing

tempo22: I will, actually I linked to the stack overflow post about it in the commit message.
https://github.com/ember-learn/ember-cli-addon-docs/pull/266/commits/381f17e9c8bf5433ff2bbac1905443291a7ce934

pzuraq: ahh gotcha, missed that. I think the comment is good anyways, thanks for this!

pzuraq:
Looks like the failures are legitimate 😕 may have been something else that changed and is breaking now though, but I'm unsure how that could be since it's the locked dependencies test...

tempo22:
Master branch is having the same test failing for me.
Should i wait for a fix on master before merging?

samselikoff:
Ran into this failure and got a fix coming. Tbh I have no idea how this test ever passed...

samselikoff:
Thank you so much for the contrib, I missed this when I pushed up 301!

PR #271, opened 2018-10-29
Fix usage of a test selector on a tagless component

anehx:
`{{docs-header/link}}` is a tagless component which means we can't apply a test selector on it. `ember-test-selectors` used to throw a warning however they changed it to an assertion which results in breaking all addons using this addon and `ember-test-selectors`. I actually don't know how those tests for this functionality ever worked :thinking:

### DIFF
--- a/addon/components/docs-header/template.hbs
+++ b/addon/components/docs-header/template.hbs
@@ -23,8 +23,8 @@
 
       {{docs-header/search-box query=query on-input=(action (mut query))}}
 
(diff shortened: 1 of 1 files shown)


dcyriller:
It sounds that `ember-test-selectors@2` and `ember-cli-addon-docs@latest` are currently incompatible 😢 , at least if you're using the `docs-header` component.

dcyriller:
The tests are failing in CI, but I can't make them fail on my local machine.

dfreeman:
#341 was just merged, which fixes this. Sorry for the long wait!

PR #277, opened 2018-11-15
fix(build): handle module unification addons

buschtoens:
I wanted to play around with ember-cli-addon-docs in a Module Unification addon, but hit some problems. This in combination with https://github.com/ember-cli/ember-cli/pull/8197 fixes all problems for me.

The `ember-fetch` upgrade is required for monorepo addons. I'll rebase this PR, when https://github.com/ember-learn/ember-cli-addon-docs/pull/284 is merged.

This PR here is also lacking tests for MU.


buschtoens:
So this is working now for classic and MU. _However_, the story for how to test an addon with classic _and_ MU dummy apps [is still shaking out](https://github.com/emberjs/ember.js/issues/16373). I guess until then, we can't really write tests for MU.

It would be sweet however, if this PR could be merged nonetheless, as the changes are rather minimal and don't interfere with the classic build pipeline, but would allow early MU adopters to keep experimenting with the mainline release of `ember-cli-addon-docs`.

buschtoens:
There is one minor issue I noticed: `ember-svg-jar` icons don't work, but this is most likely an upstream issue. I'll investigate.

buschtoens:
The `app` dir is only partially merged with the MU dummy app, meaning that primarily the `docs.api.item` route template is not made available out of the box.

The `inject-media` initializer is executed. The transitions for `shift+slash` don't seem to work, bu they are also broken on master. Try pressing `shift+slash` here: https://ember-learn.github.io/ember-cli-addon-docs/

buschtoens:
Maybe we can do some trickery with [`Route#templateName`](https://www.emberjs.com/api/ember/3.5/classes/Route/properties/templateName?anchor=templateName) to avoid the manual re-export.

samselikoff:
kind of want to wait until MU is finalized to support this? what does everyone think? 

Can always come back to it if/when that stuff ever shakes out

rwwagner90:
I've heard from those on Ember core that MU is going through a rethink and will likely differ from the original vision. With that in mind, I am going to close this for now, and we can revisit when things flesh out more.

Issue #267, opened 2018-10-16
API Docs: classes in modules


anehx:
Currently, if we have a class in a module e.g `my-addon/lib/some-class.js` which looks like this:

```js
/**
 * Some class which does some stuff..
 *
 * @class SomeClass
 */
export default class SomeClass {
  /**
   * Some method which magically always returns 42
   *
   * @method someMethod
   * @return {Number} Always 42
   */
  someMethod() {
    return 42
  }
}
```

As a user I'd expect to see this class documented the same as if it was located in `my-addon/service/some-service`. I'd expect to see the documented properties, accessors, methods etc.

What do you think about this? Is there a reason that this is implemented like this?

samselikoff:
I have some code for this, just need to get back around to it. I think we'll be able to support these "special cases".

PR #280, opened 2018-11-15
Show single exports of resolved types in nav bar

samselikoff:
If a module exports a single object, show that object under its resolved type in the side navigation.

Example from Mirage's docs: `Collection` is a single class export from the `addon/orm/collection` module.

image

Closes https://github.com/ember-learn/ember-cli-addon-docs/issues/267


samselikoff:
@pzuraq @dfreeman Could I get a quick once-over of this when y'all have the time? This makes single-export modules appear a bit nicer in the nav index

samselikoff:
> My only minor nit is that I'm not sure why NavigationIndexGenerator is a class instead of just a collection of functions since it has no state

@dfreeman good question... I don't really have a lot of experience "organizing free functions" (for lack of a better term), so I usually reach for classes since most of my work has been in OOP systems. I find them to be the easiest way to organize my thinking & refactor into small methods, and of course, there's always the chance that you'll add state later. (In Mirage I wrote the json:api graph serializer initially as a set of functions, then as a bunch of methods on a class, then added some state that tracked "has-been-serialized" models & collections, which I found drastically simplified some of the logic).

That being said I know I simply am not equipped to work on systems that are mostly comprised of functions, so those organizational patterns are just not familiar to me. I'm happy to take this opportunity to learn, if you have some pointers for me.

(One of the existing modules changed in this PR had a combination of a class export and free-floating functions [as you can see here in master](https://github.com/ember-learn/ember-cli-addon-docs/blob/master/lib/broccoli/docs-compiler.js), but I personally find this sort of file hard to grok, since every function is at the same level. I don't know which ones are "higher-level" and "lower-level". And that's one of the reasons I extracted to a class with "the important stuff at the top.")

Again we can merge this now but I figured I'd write this up because it's something I've been interested on getting a better understanding of!

PR #286, opened 2018-11-20
Fix markdown parser

gossi:
With version `0.6.0` and using angle bracket syntax inside your markdown files will actually brake the build.

This is **not** working:

```
<DocsDemo @class="docs-demo" as |demo|>
  {{#demo.example name='demo.hbs'}}
    <FooBar @label="fubar"/>
  {{/demo.example}}

  {{demo.snippet 'demo.hbs'}}
</DocsDemo>
```

This patch fixes the parser to not only handle curlies but also angle brackets :)

### DIFF
--- a/lib/utils/compile-markdown.js
+++ b/lib/utils/compile-markdown.js
@@ -47,7 +47,9 @@ function compactParagraphs(tokens) {
     }
 
     balance += count(/\{\{#/g, token.text);
+    balance += count(/<[A-Z]/g, token.text);
     balance -= count(/\{\{\//g, token.text);
+    balance -= count(/<\/[A-Z]/g, token.text);
   }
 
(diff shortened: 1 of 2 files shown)


samselikoff:
Ty!

miguelcobain:
I *think* this broke something.

When I write a paragraph like this in my `template.md`:

```markdown
Up until now the guides assumed you passed in an array of data to the `@data` argument of `<YetiTable>`.
```

the result is

screen shot 2019-01-23 at 17 43 46

However, if I change `<YetiTable>` to `YetiTable>` it works fine:

screen shot 2019-01-23 at 17 45 31

So, the opening `<` is making things look unformatted. Basically everything gets stuck inside a single `<p>`.

Issue #263, opened 2018-10-11
Make keyboard shortcuts optional


dbollinger:
In addition to the `{{docs-keyboard-shortcuts}}` component, there are some keyboard shortcuts baked into `controllers/application.js` and `docs-viewer/component.js` that are enabled on initialization.  These can interfere with components that have custom keyboard interactions.

A configuration option that disables shortcut initialization could solve this issue for some users.

PR #288, opened 2018-11-21
Updates ember-keyboard configurations to be more flexible

dbollinger:
For #263 

- Removes `addon/controllers/application.js` because it seems like it's duplicating `docs-keyboard-shortcuts`
- Updates components that use `ember-keyboard` to default the keyboardActivated property to true instead of setting in the init methods so users can optionally render those components with it set to false.
- Leaves all existing keyboard configurations enabled by default


dbollinger:
This could probably use some documentation around which components support which shortcuts

rwwagner90:
@samselikoff @pzuraq any reason not to merge this? It looks good to me.

samselikoff:
Looks good!

PR #300, opened 2019-01-19
Update ember-component-css to fix 3.6 deprecation

cibernox:
This version includes https://github.com/ebryn/ember-component-css/pull/308 which fixes this addon for Ember 3.6+


cibernox:
If after this is merged we can get a new release that would be 🔝

Issue #298, opened 2019-01-11
Sub-dependency of JSDom, ip-regex, is not compatible with Node v6


pzuraq:
A subdependency of this addon is no longer compatible with ip-regex. It's possible to work around this for the moment with `resolutions` in package.json, but it may eventually become more problematic.

It looks like the JSDom dependency was added as part of @savvymas and @ef4's migration script, which I _think_ could be removed at this point. If folks still need to update, we could point them to first updating to a previous version of `ec-addon-docs` and then to the current version.

pzuraq:
See [this issue for more details](https://github.com/ember-cli/ember-cli/issues/8360)

ef4:
There's no reason we can't make sure the migration code loads lazily, such that it has no impact on you if you aren't migrating.

pzuraq:
We would have to make the dependencies `optionalDependencies`, since the failure will occur even before install finishes if the engine conflict is detected. I think that could work.

samselikoff:
Unless someone is volunteering to take on that work that seems more complex & time consuming. I'm all for pointing folks to an earlier version as an interim upgrade step.

dfreeman:
We're in the [home stretch for Node 6](https://github.com/nodejs/Release#nodejs-release-working-group), but this is definitely painful for anyone who's still (rightly!) testing their addons against it in CI. It looks like @simonihmig opened [a PR to `tough-cookie`](https://github.com/salesforce/tough-cookie/pull/141) to constrain the `ip-regex` dependency; if that goes through soon, we may not have to take action here.

simonihmig:
Yes, but unsure how quickly this will get released... I used `resolutions` to "fix" it for now, but FWIW for my addon `jsdom` was just a devDependency...

samselikoff:
Could y'all weigh on in https://github.com/ember-learn/ember-cli-addon-docs/pull/302?

PR #301, opened 2019-01-20
Fix deploy logic for custom domains

samselikoff:


samselikoff:
should probs add some tests but this got my deploys working for http://www.ember-cli-mirage.com/versions/master/

dfreeman:
Looks like CI is failing because of #298. If we can land on a solution for that, though, I think this can supercede #266 and we can maybe cut a v0.6.2 release.

PR #302, opened 2019-01-21
Pin tough-cookie to support Node 6

samselikoff:
Should be ablve to remove this if/when https://github.com/salesforce/tough-cookie/pull/141 is merged.

Fixes https://github.com/ember-learn/ember-cli-addon-docs/issues/298

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,9 @@
     "qunit": "^2.6.2",
     "qunit-dom": "^0.8.0"
   },
+  "resolutions": {
+    "**/tough-cookie": "~2.4.0"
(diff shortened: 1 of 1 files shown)


samselikoff:
Lots of good code in the migrator script and it only adds one dep (`jsdom`) to our package.json. Seems like we should keep it around if possible. Also Node 6 being EOL'd in April & that PR on `touch-cookie` hopefully being merged means this resolutions hack can be removed soon...

Thoughts?

samselikoff:
Believe it does

PR #304, opened 2019-01-21
Use CSS Custom Property for brand color

samselikoff:
This implementation has some debt. We can decide if we want the feature now or want to figure out a cleaner solution now.

If we could just use CSS Custom Properties, this would be super clean as the Tailwind config var would just be `--brand-primary`. However, there's not full browser support, so I wanted to provide a fallback. The fallback is to define a vanilla `color: ` (or background-color or border-color) property before the variable use.

I used a Tailwind Component to insulate the rest of the codebase from this knowledge, that way we can continue to use `docs-color-brand` and have it work everywhere.

I had to use `!important` because components come before utilities, so a `docs-text-grey-dark` was overriding the brand color. If/when we get the router service in to improve the docs links in the header/nav components, this can be dropped.

### DIFF
--- a/addon/components/docs-header/link/template.hbs
+++ b/addon/components/docs-header/link/template.hbs
@@ -28,9 +28,14 @@
   {{#link-to route
     class=(concat
       'docs-px-4 docs-py-5 docs-transition docs-uppercase docs-text-xxs
-      docs-font-bold docs-no-underline docs-text-grey-darkest '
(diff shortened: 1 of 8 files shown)


samselikoff:
Docs here https://github.com/ember-learn/ember-cli-addon-docs/commit/452394a61bb52650a38d934f9e946d42b0e0f916

PR #307, opened 2019-01-25
Add docs route in quickstart section

dcyriller:
This has proved to fix a weird bug I had in an addon.

**Reproduction Steps**
- visit https://qonto.github.io/ember-phone-input/
- click `Read the docs`
- click `API REFERENCE -> Components -> phone-input`
- refresh the page

**Bug**
API section is not rendered anymore.

**Investigation**
For some reason, the `ember-cli-addon-docs/routes/docs`'s `model` hook is not played.

**Resolution / work-around**
I've created a route `tests/dummy/app/pods/docs/route.js` and extended it from e-c-a-d's. The model is now always triggered.

### DIFF
--- a/tests/dummy/app/pods/docs/quickstart/template.md
+++ b/tests/dummy/app/pods/docs/quickstart/template.md
@@ -59,6 +59,12 @@ all docs pages in your site.
     {{/docs-viewer}}
   {{/docs-snippet}}
 
+  {{#docs-snippet name='quickstart-skeleton.js' title='tests/dummy/app/routes/docs.js'}}
+    import BaseDocsRoute from 'ember-cli-addon-docs/routes/docs'
+
+    export default class DocsRoute extends BaseDocsRoute {}
+  {{/docs-snippet}}
+
 5. **Create your Markdown templates.** Markdown templates contain the actual
 documentation for your addon and live in the folder
 `tests/dummy/app/templates/docs`. Since Addon Docs supports Markdown out


samselikoff:
Believe the problem is you've overridden `docs/route.js`. Ideally we make this less brittle, but for now you shouldn't overwrite that route. 

As far as how to get the logic you currently have in your project's `docs/route.js`, you have two options:

1. If you want Installation to be at the `/docs` URL, you could define it as such in your router:

    ```js
    this.route('installation', { path: '/' })`
    ```

    Then, users clicking on Docs or visiting `/docs` would see the Installation guide.

2. If you want to keep Installation at `/docs/installation` and just redirect `/docs` there, create a `/docs/index/route.js` and add the redirect logic there:

    ```js
    // docs/index/route.js
    import Route from '@ember/routing/route';

    export default Route.extend({

      beforeModel() {
        this.transitionTo('docs.installation');
      }

    });
    ```

Let me know if you have any more questions!

dcyriller:
Okay I understand, I shouldn't have generated a file for `docs/route.js`. Understood, thank you 👍

I wonder how we could prevent other users running into this issue 🤔

samselikoff:
@dcyriller Possibly if the API stuff was wrapped into a component and you render it. But then that makes it harder to block rendering based on the data loading, so for now I think it's our best bet.

One idea would be to generate the files into the host app's project. Places more upgrade burden on host apps but makes it more explicit.

PR #308, opened 2019-01-29
Some Suggestions

noslouch:
Hey!

Love this addon. I'm just diving into the docs and saw a few things I thought could use some small fixes.

Each commit is a discrete change and should hopefully be self-explanatory. Hope this is helpful!


samselikoff:
Thank you sir!

PR #309, opened 2019-01-29
lastest and master are not in version selector on second open

noslouch:
`removeObjects` mutates the original array


samselikoff:
Thank you! ❤️

PR #312, opened 2019-01-31
support opening tags inside backticks

### LINKED ISSUE #306: using angle brackets inside backticks should be possible
I *think* https://github.com/ember-learn/ember-cli-addon-docs/pull/286 broke something.

When I write a paragraph like this in my `template.md`:

```markdown
Up until now the guides assumed you passed in an array of data to the `@data` argument of `<YetiTable>`.
```

the result is

screen shot 2019-01-23 at 17 43 46

However, if I change `<YetiTable>` to `YetiTable>` it works fine:

screen shot 2019-01-23 at 17 45 31

So, the opening `<` is making things look unformatted. Basically everything gets stuck inside a single `<p>`.

miguelcobain:
Fixes #306 

Basically I updated the regexes with negative lookbehinds and lookaheads. From my quick research, negative lookbehinds were introduced in node 6, so we should be good. lookaheads were always supported, afaik.

### DIFF
--- a/lib/utils/compile-markdown.js
+++ b/lib/utils/compile-markdown.js
@@ -46,10 +46,13 @@ function compactParagraphs(tokens) {
       last.text = `${last.text} ${token.text}`;
     }
 
-    balance += count(/(?<!`.*){{#/g, token.text);
-    balance += count(/(?<!`.*)<[A-Z]/g, token.text);
-    balance -= count(/{{\/.*}}(?!.*`)/g, token.text);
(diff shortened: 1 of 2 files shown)


miguelcobain:
Ok, as a workaround, I just removed the code between `` to test for the balance. This way, the opening tags inside backticks don't mess with the paragraph compacting algorithm.

miguelcobain:
Just confirmed that this fixed all the issues I was having on the yeti table website. 🎉

samselikoff:
Awesome! Thank you dude 👊

PR #315, opened 2019-01-31
hide snippet nav when demo has a single snippet

miguelcobain:
Only shows snippet nav when more than one snippet is present. Looks like this:

screen shot 2019-01-31 at 20 43 32

### DIFF
--- a/addon/components/docs-demo/template.hbs
+++ b/addon/components/docs-demo/template.hbs
@@ -6,7 +6,7 @@
   )}}
 
   <div>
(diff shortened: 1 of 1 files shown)


samselikoff:
looks awesome! Any chance we could add a 1px border to the top, to differentiate the two sections?

miguelcobain:
Ok, updated.
Added top border with the same background as the nav. Only added when the nav is hidden.

screen shot 2019-01-31 at 20 54 21

PR #321, opened 2019-02-12
Expose compileMarkdown as public API

### LINKED ISSUE #318: Make the markdown renderer w/syntax highlighting a public function
Copied [here](https://github.com/samselikoff/ember-cli-mirage/pull/1493/files#diff-5fa139b2336a757ef4df0e2081bfdb2b). Ideally this is an import from AD.

samselikoff:
- closes #318


samselikoff:
Use case here was rendering Markdown content in Mirage's blog ([example](http://www.ember-cli-mirage.com/blog/how-to-reliably-override-npm-dependencies-1497)).

AddonDocs sites should be able to do this using the same Markdown renderer used by AddonDocs' own components.

The implementation here moves the `compile-markdown` function into a new `addon-docs-shared` subdirectory, which is acting like its own npm package. This is so that it can be `require()`'d by node scripts, and also Auto Import'd by Ember app code. This is to support run-time use cases (like Mirage's blog).

Test included in the dummy addon.

For more history see [this Auto Import issue](https://github.com/ef4/ember-auto-import/issues/188).

PR #329, opened 2019-03-02
Allow docs-hero class to be customizable

miguelcobain:


miguelcobain:
@jessica-jordan I added the class property on `component.js` as requested.
Also passed the class argument on the docs-hero demo2. I don't think the class argument should have its own demo on that page (i.e a demo3), but let me know if you think that is needed. 👍

dfreeman:
Thanks for taking this on @miguelcobain! One thought — this solves the issue for curly components, but `<DocsHero class='...' />` will still fail. This may be a bigger issue across all of our tagless components that we should look at separately, and I don't want to unnecessarily delay this PR. 

If we include a dependency on the angle brackets polyfill, do you happen to know if that would  allow us to add `...attributes` here without blowing up during parsing in older versions of Ember?

dfreeman:
I think not having a dedicated example is fine. @jessica-jordan unless you have any objections, I'll merge this?

jessica-jordan:
Yes sure, let's do so! Just didn't merge yet, as there were some issues with CI before, but this seems :sparkles:

PR #332, opened 2019-03-04
Fix a double slash in url that causes history corruption

ming-codes:


lib/deploy/plugin.js
```diff
@@ -155,12 +155,18 @@ module.exports = class AddonDocsDeployPlugin {
...
-    let updated = contents.replace(/\/?ADDON_DOCS_ROOT_URL\/?/g, addonDocsRootURL)
-      .replace(/%22ADDON_DOCS_DEPLOY_VERSION%22/g, encodedVersion);
+    let updated = this._macroReplaceIndexContent(contents, addonDocsRootURL, encodedVersion);
 
     fs.writeFileSync(indexPath, updated);
   }
 
+  _macroReplaceIndexContent(contents, addonDocsRootURL, encodedVersion) {
+    return contents
+      .replace('%2FADDON_DOCS_ROOT_URL%2F', encodeURIComponent(addonDocsRootURL))
```
ming-codes: @samselikoff This first replace work on the `rootURL` in meta tag.

samselikoff:
Thanks!

Merging this for now, but I think we need to back out of our ADDON_DOCS_ROOT_URL abstraction. It means one build is really two, which confuses other tools (e.g. Prember) that aren't aware of this.

samselikoff:
@ming-codes Believe we might have had a regression with this PR...

@chriskrycho is trying to make a new AD site for https://typed-ember.github.io/ember-cli-typescript/versions/update-docs/

If you look at his build, you'll see this in the config:

```
{
  "modulePrefix": "dummy",
  "environment": "production",
  "rootURL": "//ember-cli-typescript/versions/update-docs//",
```

That rootURL is not right and causing his app not to load.

Any chance you could take a look?

ming-codes:
It appears they're still using the unpatched version

https://github.com/typed-ember/ember-cli-typescript/blob/06ee23e383c3e4f9238e9bfdbc092c7db68666f8/package.json#L73

`^^0.6.5` should resolve to just `0.6.5`. I would recommend all pre 1.0 usages should just use semver `< 1.0.0`

samselikoff:
@ming-codes Oh good catch! Thought they were on 0.6.6. Will tell them to upgrade 👍

ming-codes:
Just checked the link. Looks like it’s working now 👍

PR #336, opened 2019-03-09
[FEATURE] Pass doc descriptions through htmlbars

ming-codes:
This PR passes the api doc description through htmlbars compiler. This allows you to use any components that are accessible from the app. For my specific case, I was trying to pull a code snippet from elsewhere. But you can certainly get a running example in there.

@samselikoff This is a crude proof of concept PR. Let me know what you think and I can update the PR to be more robust.

### DIFF
--- a/addon/components/api/x-class/template.hbs
+++ b/addon/components/api/x-class/template.hbs
@@ -1,7 +1,9 @@
 <h1 class='docs-h1' data-test-class-name>{{class.name}}</h1>
 
 {{! wrapping in a div seems to work around https://github.com/ember-learn/ember-cli-addon-docs/issues/7 }}
-<div data-test-class-description>{{{class.description}}}</div>
+<div data-test-class-description>
+  {{api/x-description class}}
+</div>
 
 {{#if (or class.exportType hasToggles)}}
   {{#api/x-meta-panel as |panel|}}
--- a/addon/components/api/x-component/template.hbs
+++ b/addon/components/api/x-component/template.hbs
@@ -1,7 +1,9 @@
 <h1 class='docs-h1' data-test-component-name>{{component.name}}</h1>
 
 {{! wrapping in a div seems to work around https://github.com/ember-learn/ember-cli-addon-docs/issues/7 }}
-<div data-test-component-name>{{{component.description}}}</div>
+<div data-test-component-name>
+  {{api/x-description component}}
+</div>
 
 {{#if hasToggles}}
   {{#api/x-meta-panel as |panel|}}
--- a/addon/components/api/x-description/component.js
+++ b/addon/components/api/x-description/component.js
@@ -1,0 +1,27 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+
+import { createTemplateFactory } from '@ember/template-factory';
+
+import layout from './template';
+
+export default Component.extend({
+  layout,
+  tagName: '',
+
+  item: null,
+
+  htmlbars: computed('item.htmlbars', function() {
+    let htmlbars = this.get('item.htmlbars');
+
+    if (htmlbars) {
+      return createTemplateFactory(JSON.parse(htmlbars));
+    }
+
+    return null;
+  }),
+}).reopenClass({
+
+  positionalParams: [ 'item' ]
+
+});
--- a/addon/components/api/x-description/template.hbs
+++ b/addon/components/api/x-description/template.hbs
@@ -1,0 +1,7 @@
+{{#if item}}
+  {{#if item.htmlbars}}
+    {{api/x-description layout=htmlbars}}
+  {{else}}
+    {{{item.description}}}
+  {{/if}}
+{{/if}}
--- a/addon/components/api/x-section/template.hbs
+++ b/addon/components/api/x-section/template.hbs
@@ -27,7 +27,7 @@
   </h3>
 
   <p data-test-item-description>
-    {{{item.description}}}
+    {{api/x-description item}}
   </p>
 
   {{#if (or item.exportType shouldDisplayParams)}}
@@ -51,7 +51,7 @@
               <tr data-test-item-param>
                 <td><span class="docs-font-mono docs-font-bold docs-border-r docs-border-grey-light docs-pr-2">{{param.name}}</span></td>
                 <td><span class="docs-font-mono docs-border-r docs-border-grey-light docs-px-2">{{param.type}}</span></td>
-                <td class="docs-pl-2">{{param.description}}</td>
+                <td class="docs-pl-2">{{api/x-description param}}</td>
               </tr>
             {{/each}}
           </tbody>
--- a/index.js
+++ b/index.js
@@ -187,6 +187,8 @@ module.exports = {
 
   postprocessTree(type, tree) {
     let parentAddon = this.parent.findAddonByName(this.parent.name());
+    let parentHtmlbars = this.parent.findAddonByName('ember-cli-htmlbars');
+
     if (!parentAddon || type !== 'all') { return tree; }
 
     let PluginRegistry = require('./lib/models/plugin-registry');
@@ -195,6 +197,7 @@ module.exports = {
 
     let project = this.project;
     let docsTrees = [];
+    let parentHtmlbarsOptions = parentHtmlbars.htmlbarsOptions();
 
     this.addonOptions.projects.main = this.addonOptions.projects.main || generateDefaultProject(parentAddon);
 
@@ -212,7 +215,8 @@ module.exports = {
       docsTrees.push(
         new DocsCompiler(docsGenerators, {
           name: projectName === 'main' ? parentAddon.name : projectName,
-          project
+          project,
+          templateCompiler: parentHtmlbarsOptions.templateCompiler
         })
       );
     }
--- a/lib/broccoli/docs-compiler.js
+++ b/lib/broccoli/docs-compiler.js
@@ -10,13 +10,19 @@ const Serializer = require('../serializers/main');
 const compileMarkdown = require('../utils/compile-markdown');
 const NavigationIndexGenerator = require('./docs-compiler/navigation-index-generator');
 
-function compileDescriptions(objects) {
+function compileDescriptions(objects, options = {}) {
+  let { templateCompiler } = options;
+
   if (!objects) { return; }
   if (!Array.isArray(objects)) { objects = [objects]; }
 
   for (let object of objects) {
     if (object.description) {
-      object.description = compileMarkdown(object.description);
+      object.description = compileMarkdown(object.description, { targetHandlebars: true });
+
+      if (templateCompiler) {
+        object.htmlbars = templateCompiler.precompile(object.description);
+      }
     }
   }
 }
@@ -32,11 +38,13 @@ const relationships = [
   'arguments'
 ];
 
-function eachRelationship(collection, fn) {
+function eachRelationship(collection, fn, options = {}) {
+  let { templateCompiler } = options;
+
   collection.forEach((item) => {
     relationships.forEach((key) => {
       if (item[key]) {
-        fn(item[key], key);
+        fn(item[key], key, { templateCompiler });
       }
     });
   });
@@ -63,13 +71,15 @@ function removeEmptyModules(modules) {
   });
 }
 
-function compileDocDescriptions(modules) {
-  modules.forEach(compileDescriptions);
+function compileDocDescriptions(modules, options = {}) {
+  let { templateCompiler } = options;
+
+  modules.forEach(module => compileDescriptions(module, { templateCompiler }));
 
   eachRelationship(modules, (relationship) => {
-    compileDescriptions(relationship);
+    compileDescriptions(relationship, { templateCompiler });
 
-    eachRelationship(relationship, compileDescriptions);
+    eachRelationship(relationship, compileDescriptions, { templateCompiler });
   });
 }
 
@@ -83,6 +93,7 @@ module.exports = class DocsCompiler extends CachingWriter {
 
     this.name = options.name;
     this.project = options.project;
+    this.templateCompiler = options.templateCompiler;
   }
 
   build() {
@@ -100,7 +111,9 @@ module.exports = class DocsCompiler extends CachingWriter {
 
     removeHiddenDocs(modules);
     removeEmptyModules(modules);
-    compileDocDescriptions(modules);
+    compileDocDescriptions(modules, {
+      templateCompiler: this.templateCompiler
+    });
 
     let project = {
       id: name,


ming-codes:
PR Updated with better solution

samselikoff:
Hm, could you drop an example use or test case so I can see what end-user API looks like?

ming-codes:
It's live here https://ming-codes.github.io/more-ember-test-helpers/versions/master/docs/api/modules/more-ember-test-helpers/adapter

The code examples are linked from actual test cases via `{{docs-snippet}}`.

samselikoff:
Can you link in the source I mean?

ming-codes:
Here's the usage in source https://github.com/ming-codes/more-ember-test-helpers/blob/master/addon/adapter/index.js#L26

samselikoff:
Only thing that concerns me about this is it makes your comments non-portable. There’s something nice/possibly important about portability of markdown-only comments. 

Cc @dfreeman @pzuraq thoughts?

ming-codes:
Can you help me understand what you mean by portable? How would you port the doc?

samselikoff:
Just that

```js
/**
 * Invokes the findMany hook on the target adapter
 * under test and returns an object that represents
 * the sent request.
 *
 * {{docs-snippet name="adapter.findMany"}}
 *
 * @function findMany
 * @param {String} modelName
 *   The model name for the matching adapter that is
 *   under test.
 * @param {any[]} ids
 *   The ids of the model objects to lookup.
 * @return {AdapterRequest}
 */
```

isn't as portable a code comment as

```js
/**
 * Invokes the findMany hook on the target adapter
 * under test and returns an object that represents
 * the sent request.
 *
 * ```js
 * test('findMany', function(assert) {
 *   let request = adapter.findMany('user', [ 1, 2, 3 ]);
 * 
 *   assert.equal(request.method, 'GET', 'method');
 *   assert.equal(decodeURIComponent(request.url.search), '?ids[]=1&ids[]=2&ids[]=3');
 * });
 * ```
 *
 * @function findMany
 * @param {String} modelName
 *   The model name for the matching adapter that is
 *   under test.
 * @param {any[]} ids
 *   The ids of the model objects to lookup.
 * @return {AdapterRequest}
 */
```

is, because `{{docs-snippet name="adapter.findMany"}}` is meaningless outside the context of AddonDocs.

ming-codes:
Yes, I agree that would be more portable.

Allowing `htmlbars` in the description does not stop authors from writing portable docs, but it would arm the authors with more tools to write better docs. Isn't that the goal of this addon?

samselikoff:
I see, so the idea in this case is that your tests really are kinda your documentation, so its easiest to be able to reuse them in your API doc comments?

ming-codes:
Yes! More than that, this feature would allow the usage of `{{link-to}}` to reference other parts of the doc. I know the official Ember doc does a similar thing where they reference the guide from within the API doc.

samselikoff:
I'm still having an aversion to using something like `{{link-to}}`, because ideally if someone could use Markdown links in their comments to reference other parts of the site (and we'd be able to make them a client-side transition if they are on the same domain.)

I think I'm okay with this change but do think we should encourage folks to use Markdown as much as possible. I've moved projects across documentation systems and keeping idiosyncratic things out of doc comments is worth encouraging, imo.

I'd like to hear from @pzuraq or @dfreeman here before merging! Also, could we get a single test, perhaps adding a `{{docs-link}}` to the Sandbox and testing that?

dfreeman:
I have kind of ambivalent feelings on this. On one side, having feature parity between the markdown in your templates and the markdown in your API docs feels really nice. On the other, this removes our ability to build any kind of cross-library API docs in the future.

While we aren't doing this today, one nice feature of the "API docs as standalone data" format that we use is that in principle you could build a layer on top of a set of AddonDocs apps (e.g. for all your internal addons at work, or even for all public addons with an AddonDocs site) that allows for searching and navigation across all of them. By baking the compiled template format for a particular Ember version into the docs JSON, you can no longer render them in a different version of Ember, so the docs become bound to the specific app they're published with.

It may be that that's a totally reasonable tradeoff to make here, and that the API doc data is the wrong level of granularity for building a meta-site at anyway. While we've talked about the possibility basically as long as AddonDocs has existed, no one (that I know of) has gone too far down the path of figuring out what that would look like and how it could be usefully consumed.

samselikoff:
Great points. 

1. Can we make it so relative markdown links work as in-app transitions?

2. Are there any extension points we can think of offering so that even if we want to discourage this pattern in general, someone in Ming’s position could opt-in to this extra behavior?

ming-codes:
> On one side, having feature parity between the markdown in your templates and the markdown in your API docs feels really nice.

Yes! That was my starting angle. As a user, I tried adding curly components to the API doc and expected it to work thinking that it's rendered in the same way as markdown. I was surprised that the component is rendered as text. I tend to think this is the kind of feature that users would expect it to "just work".

> ...cross-library API docs in the future

What is this cross-library doc idea? If we make the feature something that users can opt-in if desired (maybe through `ember-cli-build.js` options), it should still be able to do cross-library doc.

It may be a good idea to put this under a feature flag so we can iterate over it.

> By baking the compiled template format for a particular Ember version into the docs JSON, you can no longer render them in a different version of Ember, so the docs become bound to the specific app they're published with.

I may need to check my understanding but I see the published doc is always bundled with the Ember version that it uses (as part of `/versions/v0.4.14/assets/vendor-*.js`). So this is really a non-issue.

If it really is an issue, we can always ship the compiler with the doc app and compile the template on client side. This will ensure version compatibility.

dfreeman:
> I may need to check my understanding but I see the published doc is always bundled with the Ember version that it uses (as part of /versions/v0.4.14/assets/vendor-*.js). So this is really a non-issue.

@ming-codes The data that powers the API docs themselves is published as a vanilla `.json` file that we fetch when the docs app boots up ([example from our own docs site](https://ember-learn.github.io/ember-cli-addon-docs/docs/ember-cli-addon-docs.json)).

You could imagine grabbing those JSON files from multiple different addons and presenting them all together in a unified site, like [docs.rs](https://docs.rs) or [rubydoc.info](https://www.rubydoc.info/gems). For that to work, though, the documentation data itself needs to avoid being bound to a particular Ember version.

> If it really is an issue, we can always ship the compiler with the doc app and compile the template on client side. This will ensure version compatibility.

We used to include the template compiler on the client side and ended up having some issues with it. Even if we solve those, however, including live code in the API docs still makes it effectively impossible to build an aggregate documentation app, since it would require that app to have every single addon installed in order to be able to run their examples.

Today, rendering the API docs doesn't even necessarily require Ember or even JavaScript—since it's "just" JSON:API data, you could use the data to render static HTML with PHP if you wanted.

> 1. Can we make it so relative markdown links work as in-app transitions?

@samselikoff I think with `href-to` installed, regular links should mostly "just work"? I haven't tried, though, and I'm not sure about relative links (which you'd want, unless we end up doing some smart `rootURL` processing on the markdown).

I feel reasonably confident we could find a way to make markdown links work nicely, though.

> 2. Are there any extension points we can think of offering so that even if we want to discourage this pattern in general, someone in Ming’s position could opt-in to this extra behavior?

We could probably come up with something, but I think it all sort of comes back to the same issue: anyone who used an extension point like this would be opting out of having their documentation be consumable by anything outside that specific instance of their docs app.

@ming-codes I want to be totally clear that I see **a ton of value** in what you're proposing, and having more dynamic API doc descriptions could improve the experience for both the doc authors and consumers. I just can't come up with an obvious way to support both live documentation _and_ an aggregate API doc browser, since we'd never be able to show the former in the latter.

@samselikoff @pzuraq I may be completely missing a way to have our cake and eat it too on this one, so if you've got any thoughts I'd love to hear them. Otherwise we may just need to make a call one way or the other on whether we think the metasite stuff is something it's important to leave the door open for.

samselikoff:
Thinking out loud... in Embroider there is a "build target" version of addons, where all "Emberisms" are compiled out. By the time Addons are in that phase, all thats left is HTML, CSS and JavaScript.

In Ming's use case, 

```
/**
 * {{docs-snippet name="adapter.findAll"}}
 *
 * @function findAll
 * @param {String} modelName
 * @return {AdapterRequest}
 */
```

there's no reason `{{docs-snippet}}` needs to execute at run-time. If it were executed at build-time, and his addon was published to NPM in that target format, the code comments in his source wouldn't depend on the Ember version. This component invocation would be replaced with the snippet at that point.

That way in the future, if we do end up making an aggregator, we would pull all addons from NPM rather than the source on GitHub. And that would give addon authors an abstraction point to handle exactly these sorts of situations.

I believe this idea is still new but does seem core to Embroider, which does seem like it has plenty of momentum, so I suppose we'd be making a bet that this pattern solidifies before the time we'd want to make that aggregation site. Or, we could just tell folks that we'll be pulling from NPM, so they have an opportunity to run some build-time code first. That way we can impose stronger constraints about what we expect to find there - i.e. nothing that depends on any particular run-time dependencies.

What do you think?

dfreeman:
If we only want to handle `{{docs-snippet}}`, that totally works, but it doesn't solve the broader "use arbitrary components" scenario @ming-codes called out in the original PR description.

I also have some concern about using `{{docs-snipet}}` in particular inline in API documentation. One of the absolutely fantastic things about inline API docs is that they can also be picked up by editors. If we start abstracting example snippets out of those docs in a way that they're hidden from all systems other than our own, I worry some that we start to to improve the authoring experience for the docs site at the expense of real-world usability of those docs in any other context.

The example function is contrived, but compare having these two tooltips pop up when you're exploring a new library:

image

image

samselikoff:
> at the expense of real-world usability of those docs in any other context

But again, if the consumption point ends up being NPM rather than GitHub (which I think is the goal), could we not compile away `{{docs-snippet}}` usage at build-time?

The idea is that `{{docs-snippet}}` would never end up in the package source on NPM.

dfreeman:
Sorry, I think I misunderstood what you were proposing before. My impression was that you were suggesting the compiled docs JSON be published to npm, and that we could preprocess that data before writing it.

You're saying we would rewrite the actual source files? I'm sure we could make that work, but it would make me a little nervous. Since we support arbitrary documentation formats in arbitrary languages via the plugin system, it seems like it would be easy to break code if we do a straight find/replace (e.g. what happens if the snippet content has `*/` in it, how do we handle newlines, etc?).

Or would it be the responsibility of the doc plugins to do the rewriting? The plugins themselves would have a better chance at knowing how to safely interpolate a snippet, but we'd be adding a whole new read/parse/generate/write postprocessing cycle per plugin to the build, and it would raise the barrier to building new plugins pretty high. I'm really not trying to rain on the idea here, just trying to think through what it would look like to build it.

> I see, so the idea in this case is that your tests really are kinda your documentation, so its easiest to be able to reuse them in your API doc comments?

If this is the biggest motivating factor, could we flip the problem on its head? What if we could generate tests from your doc comments to ensure they run cleanly?

I don't know if it would end up being a good fit for our JS-focused world, but the Rust community [has really embraced the idea](https://doc.rust-lang.org/rustdoc/documentation-tests.html). Out-of-the-box, all code in doc comments is considered part of your test suite by the tooling, and it's normal to see assertions in code snippets as a way of communicating to both the reader and the test infra what the expected state of the world is.

samselikoff:
> What if we could generate tests from your doc comments to ensure they run cleanly?

I think this is a great idea that we could definitely explore 👍

---

Back to the build-time thing. I'm saying yes, rewrite the actual source files, so what gets published to NPM is different from what is in the repo source on GitHub. I believe this is the plan for Embroider, so that all Emberisms are compiled away in the GitHub -> NPM build, leaving the source code that ends up in host apps much more portable.

**Step 1: Author and push to GitHub**

```js
// some-file.js

/**
  Here's how you use it:

  {{docs-snippet name="adapter.findAll"}}

  @function findAll
  @param {String} modelName
  @return {AdapterRequest}
*/
export default function() {}
```

**Step 2: Run npm publish**

This step involves some build-time transformation. It would be as if our Markdown Template compiler thing ran here instead of during build-time of the host app. Or if Ed's `//BEGIN-SNIPPET` thing ran as well. That way, what ends up on npm is just HTML/CSS/JS linked together with static ES6 imports. No arbitrary processing of files, adding things to trees, etc.

So in Ming's case, there would need to be some Broccoli build-time code that would get run, such that when he ran `npm publish` the above was first turned into

```js
// some-file.js

/**
  Here's how you use it:

  ``js
  test('findAll', function(assert) {
    let request = adapter.findAll('user');

    assert.equal(request.url.pathname, '/users', 'url.pathname');
    assert.equal(request.method, 'GET', 'method');
  });
  ``

  @function findAll
  @param {String} modelName
  @return {AdapterRequest}
*/
export default function() {}
```

and _that's_ the file that ended up on NPM.

**Step 3: Someone installs the package**

Now when someone else installs the package, they pull the above code from step 2 into their project, so their tooling works as expected because all dynamic Emberisms have been compiled out.

---

Just wanted to write that all out to make sure we were on the same page.

So, there's definitely more code to write because `{{docs-snippet}}` here would have to be a build-time component that ran. This is also something we can punt on from this library's perspective, but we could grant that folks who wanted to go down this path could do it. So, if Ming (or anyone else) wanted to do this, we should just figure out if it's going to cause any problems with our future plans, and if not give them our blessing, so to speak.

For the short-term I think your suggestion of generating tests from docs is going to be easiest path forward for Ming. However I do think the abstraction layer between Step 1 and Step 2 is powerful and may enable us to do more cool stuff in the future.

dfreeman:
> > What if we could generate tests from your doc comments to ensure they run cleanly?
>
> I think this is a great idea that we could definitely explore

Just found out via Twitter that our friends at LinkedIn [might be one step ahead of us](https://github.com/SparshithNR/doc-tester) on that 🙂

> Just wanted to write that all out to make sure we were on the same page

Thanks for taking the time to write that out! Definitely up to speed now on the flow you're proposing (and it's a huge part of what I love about the Embroider proposal), I was just nervous about trying to do it specifically for doc comments, specifically as part of addon-docs.

I'm fully on board with that idea in the abstract, and I could imagine a more general piece of tooling for snippet extraction + replacement. Something like that could expose a way to look up snippets by key for consumers like us when building the docs JSON, and also deal with the process of swapping out placeholders elsewhere in the build.

samselikoff:
> Just found out via Twitter that our friends at LinkedIn might be one step ahead of us on that 🙂

Woah... that is very cool! Let's keep our eyes on that for sure.

>  I was just nervous about trying to do it specifically for doc comments, specifically as part of addon-docs
> I'm fully on board with that idea in the abstract

Awesome, think we're in agreement then 👍. Definitely think a general purpose tool could end up being built & useful to AddonDocs users.

For now, @ming-codes I think your best path forward is going to be (a) pave some cowpaths and try to do a publish-time swap on your own, or (b) do what Dan suggested and try to pull your tests from your comments (instead of the reverse). That also means we're not going to accept this PR feature because we don't want comments to have any _runtime_ dependencies.

Does that all make sense? (I'm quite sure you got more than you bargained for when you first opened this issue ^_^)

rwwagner90:
@samselikoff @dfreeman it sounds like we want to close this PR. Should we go ahead and close it?

samselikoff:
Yep, thanks for circling back 👍

PR #342, opened 2019-03-25
support angle brackets with docs-demo component

noslouch:
The following snippet wasn't getting picked up by the original regex:
```hbs
<DocsDemo as |demo|>
  <demo.example @name='button-example.hbs'>
    <MyButton>
      Click Me
    </MyButton>
  </demo.example>

  <demo.snippet @name='button-example.hbs'/>
</DocsDemo>
```

This shouldn't break existing classic component invocation of `docs-demo`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,8 +84,8 @@ module.exports = {
     includer.options.includeFileExtensionInSnippetNames = includer.options.includeFileExtensionInSnippetNames || false;
     includer.options.snippetSearchPaths = includer.options.snippetSearchPaths || ['tests/dummy/app'];
     includer.options.snippetRegexes = Object.assign({}, {
-      begin: /{{#(?:docs-snippet|demo.example)\sname=(?:"|')(\S+)(?:"|')/,
-      end: /{{\/(?:docs-snippet|demo.example)}}/,
+      begin: /(?:{{#|<)(?:docs-snippet|demo.example)\s@?name=(?:"|')(\S+)(?:"|')/,
+      end: /(?:{{|<)\/(?:docs-snippet|demo.example)(?:}}|>)/,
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -84,8 +84,8 @@ module.exports = {
     includer.options.includeFileExtensionInSnippetNames = includer.options.includeFileExtensionInSnippetNames || false;
     includer.options.snippetSearchPaths = includer.options.snippetSearchPaths || ['tests/dummy/app'];
     includer.options.snippetRegexes = Object.assign({}, {
-      begin: /{{#(?:docs-snippet|demo.example)\sname=(?:"|')(\S+)(?:"|')/,
-      end: /{{\/(?:docs-snippet|demo.example)}}/,
+      begin: /(?:{{#|<)(?:docs-snippet|demo.example)\s@?name=(?:"|')(\S+)(?:"|')/,
+      end: /(?:{{|<)\/(?:docs-snippet|demo.example)(?:}}|>)/,
```
dfreeman: For this to work for standalone snippets, we'd need to accept `DocsSnippet` here as well, wouldn't we?

noslouch: Ah, yeah that’s right. Good catch.

rwwagner90:
@noslouch @dfreeman where are we on this?

noslouch:
Nowhere! I’m not sure where to start with a test for this 😞

dfreeman:
@noslouch I think what I'd do is add a route in the testing addon (under `test-apps/new-addon`) that has snippets in each of the formats we want to support and then write an acceptance test that navigates to that page and verifies the expected content shows up for each.

rwwagner90:
@noslouch do you think you could maybe add the tests @dfreeman described?

PR #345, opened 2019-03-30
Upgrade ember-concurrency

josemarluedke:
This should help us fixing some issues when using ember canary.


dfreeman:
It looks like @pzuraq made the same fix and released it as 0.6.8, so I'm going to go ahead and close this out. Thanks for the PR @josemarluedke!

pzuraq:
Ah, my bad, didn't see this!

PR #349, opened 2019-04-04
Converts node tests to use mocha instead of qunit.

scalvert:
As briefly discussed in https://github.com/ember-learn/ember-cli-addon-docs/pull/343#issuecomment-479818068, moving to mocha as the test library for node tests will allow us to take advantage of other packages (`ember-cli-*` utils, for instance) that internally use mocha. This opens the door to leveraging more prior art in writing more robust tests for things like blueprints, etc.

This PR converts all tests in `tests-node` to use `mocha`/`chai`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -69,10 +69,16 @@ module.exports = {
 
     // node test files
     {
+      globals: {
+        describe: true,
+        it: true,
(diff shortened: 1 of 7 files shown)


dfreeman:
That was fast 😁 — thanks @scalvert!

PR #355, opened 2019-04-09
feat: Better nav for generic modules

### LINKED ISSUE #353: Use module name only (drop path) for API Modules nav
Here's the Modules section of the nav for Ember Animated:

image

 and Ember Decorators:

image

I know they're not unique, but what do y'all think of dropping the path from the nav bar?

Here's what it would look like for Ember Animated:

image

Seems much better! And in the event an Addon exports two modules with the same name, we can figure that out when we hit it? (I have some ideas but in the meantime, I can't find an example of this in the wild).

samselikoff:
Closes #353


samselikoff:
AddonDocs:

image

AddonDocs Sandbox:

image

Ember Animated:

image

dfreeman:
This looks great! I think the one thing the jumps out at me is it feels a little unexpected to have an `@` at the top level even when it's not a package scope.

samselikoff:
I added that due to Chris' comment in https://github.com/ember-learn/ember-cli-addon-docs/issues/353...

but now I'm realizing ember-decorators is actually a package scope?

Let me try this with ember-decorators and see what happens

pzuraq:
Yeah, `@ember-decorators` is the scope where folks actually import things, I think the sidebar should always reflect that, whatever it is 😄 this could be tricky for apps, but I don't think we're supporting them just yet?

samselikoff:
image

😂 looks like I can remove that extra `@`!

addon/components/docs-viewer/x-autogenerated-api-docs/component.js
```diff
@@ -0,0 +1,102 @@
...
+          children: [
+            { name: 'compile-markdown', children: [], path: "modules/ember-cli-addon-docs/utils/compile-markdown"  },
+          ]
+        },
+      ]
+    };
+    ```
+  */
+  moduleIndex: computed(function() {
+    let modules = this.get('sections').filter(section => section.type === 'modules')[0].items;
```
josemarluedke: Just upgraded to latest and started hitting the following issue in this line:

Screen Shot 2019-04-09 at 3 05 19 PM

I don't have any modules for some reason.

samselikoff: @josemarluedke woops! Will fix soon, thanks for the report

samselikoff: @josemarluedke Try 6.10

josemarluedke: Yep, that's fixed in 6.10. Thank you.

PR #368, opened 2019-05-24
Update to Ember & Ember-cli 3.10

cibernox:


cibernox:
@samselikoff I'm not sure what is going on, but I can't find a way to tell eslint to ignore `test-apps/new-addon/node_modules/.bin/sha.js`. I tried the .eslintrc.js in the root, the one in `test-apps/new-addon`, the `.eslintignore`... No matter what I do it seems to lint that file.

samselikoff:
how bizarre...

I wonder why it's only happening for versioned tests?

samselikoff:
Also we should get dependabot up in here

cibernox:
It happens with the versioned tests because those are then ones running `eslint .`
It's easy to reproduce locally if you can give it a go.

cibernox:
@samselikoff I discovered that is becase eslint is following a symlink that goes to the root's node_modules/.bin. 
I found reports of inconsistent behavior of eslintignore with symlinks, so I took a more... *unorthodox* approach: https://github.com/ember-learn/ember-cli-addon-docs/pull/368/files#diff-d65fb4cb059197c04e3e516a2b216050R13

cibernox:
@samselikoff finally green

samselikoff:
TYVM sir!

.travis.yml
```diff
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "6"
+  - "8"
```
Turbo87: @samselikoff `package.json` still claims that this project is compatible with Node 6. was this an intentional change?

.travis.yml
```diff
@@ -52,9 +52,8 @@ jobs:
       script: yarn test:test-apps
 
     - stage: versioned tests
-      env: EMBER_TRY_SCENARIO=ember-lts-2.8
```
Turbo87: dropping support for Ember 2.8 could also be considered a breaking change 🤔

.travis.yml
cibernox: Isn't this part of the default blueprint now?

.travis.yml
cibernox: Same, this was generated by the default blueprint. Supporting the last 2 LTS releases seems of the 2.X cycle seems very reasonable tho.

.travis.yml
Turbo87: yes, and in https://blog.emberjs.com/2019/05/21/ember-3-10-released.html we specifically recommended to release a new major version when dropping support for Node 6

.travis.yml
Turbo87: it is reasonable, yes. but dropping support should still be considered a breaking change and needs a major version bump then.

PR #382, opened 2019-06-12
Add showImportPaths option

samselikoff:


samselikoff:
I could use some eyes [here](https://github.com/ember-learn/ember-cli-addon-docs/pull/382/files#diff-168726dbe96b3ce427e7fedce31bb0bcR57), I tried to use `findImporter` and grab the config from there, but it wasn't working. When I added a `console.log(this.app)` to that line, it was hit 6 times. The first 5 times it was defined and the 6th it was undefined. So I wrapped this all in a conditional 🤷‍♂ 

I'm sure there's a better way to do this! The config I tested against is `ember-cli-mirage`, I added the option to its `ember-cli-build.js` file like this:

```js
'use strict';

const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');

module.exports = function(defaults) {
  let app = new EmberAddon(defaults, {
    'ember-cli-babel': {
      includePolyfill: true
    },
    'ember-cli-addon-docs': {
      showImportPaths: false
    },
    'ember-cli-addon-docs-esdoc': {
      packages: [ 'ember-cli-mirage', '@miragejs/server' ]
    }
  });

  /*
    This build file specifies the options for the dummy test app of this addon,
    located in `/tests/dummy`. This build file does *not* influence how the addon
    or the app using it behave. You most likely want to be modifying `./index.js`
    or app's build file.
  */

  return app.toTree();
};
```

samselikoff:
Nvm – switched from using `ember-cli-build.js` in the host app to `tests/dummy/config/environment.js` and the existing config stuff worked!

PR #390, opened 2019-06-28
When following the quickstart of onboarding, launching app fails due to missing //

AwakenedClaymore:
Need to add the missing // so we can start the app and continue the quickstart

### DIFF
--- a/blueprints/ember-cli-addon-docs/index.js
+++ b/blueprints/ember-cli-addon-docs/index.js
@@ -30,7 +30,7 @@ module.exports = {
       configContents = configContents.replace(/([ \t]+)if \(environment === 'production'\) {/, [
         '$&',
         '$1  // Allow ember-cli-addon-docs to update the rootURL in compiled assets',
-        '$1  ENV.rootURL = \'ADDON_DOCS_ROOT_URL\';'
+        '$1  // ENV.rootURL = \'ADDON_DOCS_ROOT_URL\';'
       ].join('\n'));
 
       if (configContents.indexOf('ADDON_DOCS_ROOT_URL') === -1) {


dfreeman:
Thanks for the PR, @AwakenedClaymore — why does there need to be a `//` there?

AwakenedClaymore:
@dfreeman  -- When I try to start the app after onboarded this, I got an error for it. 
My environment.js file  shows this
`  // if (environment === 'production') {
   // Allow ember-cli-addon-docs to update the rootURL in compiled assets
   ENV.rootURL = 'ADDON_DOCS_ROOT_URL';}`

and then I get this 
`Unexpected token ;`

So if I add a comment to the last line, it will work.

Unless, this is too specific to my project?

dfreeman:
When I run the generator in a fresh addon, I see this code, which looks correct to me:

```js
  if (environment === 'production') {
    // Allow ember-cli-addon-docs to update the rootURL in compiled assets
    ENV.rootURL = 'ADDON_DOCS_ROOT_URL';
    // here you can enable a production-specific feature
  }
```

I can't quite tell from the snippet you posted where the newlines are for you?

AwakenedClaymore:
I see why. My app had this before bringing in the addon
`  // if (environment === 'production') {}`

so then after bringing in the addon, it becomes

`// if (environment === 'production') { // Allow ember-cli-addon-docs to update the rootURL in compiled assets ENV.rootURL = 'ADDON_DOCS_ROOT_URL';}`

Looking at it again, I don't need this part in my app because its a UI library it gets consumed by others, so it will never go into `environment === 'production'` 

Closing.

PR #393, opened 2019-07-09
Update quick start template

barryofguilder:
This updates the "Quick Start" template to use the `demo-container` class instead of just `container`.


samselikoff:
Thanks. `.container` is meant to be an example, and the `docs-*` classes are meant to be private... I guess we could add the CSS needed for a container class here. It's a bit confusing though.

rwwagner90:
I don't think we should publicly advertise the `docs-` classes, so I am in favor of closing. That okay with you @samselikoff?

samselikoff:
Yeah we can close. Should maybe add an issue to document "how to add custom styles to an AddonDocs" which really just amounts to saying, write css however you want!

PR #394, opened 2019-07-10
Use "Escape" key to dismiss Keyboard Shortcuts modal

bantic:
Adds a keyUp listener for 'Escape' that dismisses the shortcuts modal
when invoked.

Also fixes an unrelated inconsistency in quote style in the
keyboard-shortcuts component.

<kbd>Esc</kbd> seems to be a common shortcut for other modals (e.g. [bootstrap](https://getbootstrap.com/docs/4.0/components/modal/#options) and [ember-modal-dialog](https://github.com/yapplabs/ember-modal-dialog#keyboard-shortcuts)).

### DIFF
--- a/addon/components/docs-keyboard-shortcuts/component.js
+++ b/addon/components/docs-keyboard-shortcuts/component.js
@@ -1,6 +1,6 @@
 import Component from '@ember/component';
 import { on } from '@ember/object/evented';
-import { later } from "@ember/runloop";
+import { later } from '@ember/runloop';
 import layout from './template';
 import { EKMixin, keyUp } from 'ember-keyboard';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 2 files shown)


addon/components/docs-keyboard-shortcuts/component.js
```diff
@@ -55,6 +55,12 @@ export default Component.extend(EKMixin, {
     }
   }),
 
+  hideKeyboardShortcuts: on(keyUp('Escape'), function() {
```
rwwagner90: Will this only trigger for this component or would this interfere with other `esc` listeners?

bantic: I'm not very familiar with ember-keyboard, but based on a read-through of their ["priority" docs](https://patience-tema-baron.github.io/ember-keyboard/#/priority), my understanding is that this function will be called any time <kbd>Esc</kbd> is pressed while a) this `{{docs-keyboard-shortcuts}}` component is rendered and b) this component's `keyboardActivated` prop is `true`.

The ember-cli-addon-docs docs suggest rendering `{{docs-keyboard-shortcuts}}` in the `application.hbs`, and its `keyboardActivated` prop is set to `true` in its `init`, so for anyone who opts in to keyboard shortcuts it seems like yes, this callback function will be called on any docs page when <kbd>Esc</kbd> is pressed.

This is the case with all the other shortcuts already, though, and (again based on my reading of ember-keyboard docs) the keyup event's propagation won't be stopped, so it seems unlikely that it would interfere with any other keyup listeners on the page.

rwwagner90: @bantic gotcha, this seems fine to me then

PR #296, opened 2019-01-07
Update ember-cli to 3.6.1

thisduck:
This PR updates `ember-cli` to 3.6.1, along with an `ember-concurrency` update.

PR #401, opened 2019-08-17
Ember 3.12, bump some deps

rwwagner90:
Closes #296 

This also bumps some deps, which hopefully fixes test failures in beta and canary.


rwwagner90:
Still getting failures on beta and canary, not sure why. ember-try with beta locally passes.

PR #403, opened 2019-09-15
fix: scroll hash anchor into view

### LINKED ISSUE #228: Linking to anchors does not scroll to point on page
Say you copy a link to a specific anchor on a page, such as:

https://ember-decorators.github.io/ember-decorators/latest/docs/api/modules/@ember-decorators/service#service

The user would expect that loading this URL in a new tab would automatically scroll you to the `#service` anchor on the page. But it doesn't seem to; it seems to keep me at the top of the page. I assume this is because of the way the page loads, how that anchor is not present at first.

If possible, it could be nice to do some JS magic to scroll down to that appropriate point on the page once the page is loaded!

makepanic:
fixes #228

### DIFF
--- a/addon/initializers/route-anchor-jump.js
+++ b/addon/initializers/route-anchor-jump.js
@@ -1,0 +1,24 @@
+import Route from '@ember/routing/route';
+import { scheduleOnce } from '@ember/runloop';
+
+export function initialize() {
+  Route.reopen({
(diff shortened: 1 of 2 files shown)


makepanic:
weird `ember s` in the project is saying that the port is already in use :thinking:

PR #404, opened 2019-09-23
Be more explicit on where label is located

jacojoubert:
This fixes an issue where if you have a component called `<Label />` it would get rendered instead of the label of the link.


addon/components/docs-viewer/x-nav-item/template.hbs
```diff
@@ -5,15 +5,15 @@
       activeClass='docs-text-brand docs-font-medium'
       data-test-id='nav-item'
       data-test-label=label}}
-      {{label}}
+      {{this.label}}
```
rwwagner90: Looking at the component, `label` is a positional param, so I think if this is a passed in thing, it should be `@label` rather than `this.label`

PR #409, opened 2019-10-10
Support standalone docs apps

### LINKED ISSUE #383: Enforce a hard boundary between addon and docs
I think we really need a hard boundary between the addon and the addon's docs site.

Addon development is different from app development because apps can choose when and how to deal with deprecations and compatibility problems, whereas (well-maintained) addons need to be very aggressive about those things. Addons generally need to be the first to update to the latest and greatest, before apps can follow. This creates a chicken-and-egg problem for AddonDocs, because it's an app that blocks addon development.

There is also the problem of AddonDocs altering the environment that your addon will be tested in. Do your addon's components accidentally depend on Tailwind's CSS reset? How would you test your addon under a new Ember optional feature that isn't supported by AddonDocs yet? Does your addon accidentally depend on helpers that are only present because AddonDocs uses them?

All of this leads me to the conclusion that we need to split the AddonDocs site into a standalone Ember app.

From a package management standpoint, this is cleanest if you use a monorepo. But there are two major downsides to monorepos that I think mean we can't force them upon everyone:

 - NPM is incapable of depending on an exact commit of a package in a monorepo (people have been complaining about this [since 2012](https://github.com/npm/npm/issues/2974))
 - hiding your addon down inside a package is definitely confusing when first approached. There is an aesthetic objection that's not wrong.

So instead we probably need to put a separate app in a `/docs` subdirectory. That app has its own package.json and yarn.lock, completely separate from the containing addon. While that gives you one more thing to update, it does come with benefits, such as letting you put off upgrading your docs site to a newer ember version even though you've already updated your addon itself to support and test under that version.

In `/docs/package.json`, if the author has chosen to use `yarn` we can use a `link:` dependency on the addon itself, and then everything Just Works. If the author is using NPM, we can use a `file:` dependency, which works correctly but won't give you live updating during development (because it make a copy of the addon, rather than linking to it). In that case we could provide a post-install step that creates a link instead.

It would probably not be a big deal to also support the monorepo use case, for people who actually want that. I don't think it's worth adding complexity for, but if we design things correctly it's not actually more work. AddonDocs just needs to accept an argument for "what is the path to the actual addon you are documenting" so it can find source code to generate API docs from, etc.

ef4:
This makes it possible to separate your docs application from your addon's dummy app. Fixes #383.

The way it works is:
 - you're still required to keep both the addon and the docs app in the same repo (so we don't need to worry about separately versioning each)
 - but they can be peers or subdirectories of each other
 - in your docs application you add ember-cli-addon-docs
 - in your docs application you must set the `documentingAddonAt` option to point at the path of the actual addon. This option is what opts you into the new behavior, otherwise nothing should be changed.

As a side-effect of refactoring the way "Edit this page" links work, I think I also fixed them for API-docs pages in all addon docs sites (not just ones using this new architecture).

### DIFF
--- a/addon/adapters/-addon-docs.js
+++ b/addon/adapters/-addon-docs.js
@@ -1,5 +1,5 @@
 import DS from 'ember-data';
-import config from 'dummy/config/environment';
+import config from 'ember-get-config';
 import { inject as service } from '@ember/service';
 
 export default DS.Adapter.extend({
--- a/addon/components/docs-viewer/x-main/component.js
+++ b/addon/components/docs-viewer/x-main/component.js
@@ -89,7 +89,7 @@ export default Component.extend({
     if (path === 'docs/api/item') {
       let { projectName } = config['ember-cli-addon-docs'];
       let model = getOwner(this).lookup('route:application').modelFor('docs.api.item');
-      let filename = model.file.replace(new RegExp(`^${projectName}/`), '');
+      let filename = model.get('file').replace(new RegExp(`^${projectName}/`), '');
       let file = addonFiles.find(f => f.match(filename));
       if (file) {
         return { file, inTree: 'addon' };
--- a/addon/components/docs-viewer/x-main/component.js
+++ b/addon/components/docs-viewer/x-main/component.js
@@ -4,7 +4,7 @@ import { bind } from '@ember/runloop';
 import { computed } from '@ember/object';
 import appFiles from 'ember-cli-addon-docs/app-files';
 import addonFiles from 'ember-cli-addon-docs/addon-files';
-import config from 'dummy/config/environment';
+import config from 'ember-get-config';
 import { getOwner } from '@ember/application';
 
 import layout from './template';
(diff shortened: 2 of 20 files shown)


addon/adapters/-addon-docs.js
```diff
@@ -1,5 +1,5 @@
 import DS from 'ember-data';
-import config from 'dummy/config/environment';
+import config from 'ember-get-config';
```
ef4: This and the many other edits like it are because we don't want to hard-code the name "dummy" as our app's modulePrefix.

addon/components/docs-viewer/x-main/component.js
```diff
@@ -70,26 +70,38 @@ export default Component.extend({
...
     }
 
-    path = path.replace(/\./g, '/');
-
-    let { projectHref, projectPathInRepo, primaryBranch } = config['ember-cli-addon-docs'];
-    let projectPath = projectPathInRepo ? `/${projectPathInRepo}/` : '/';
-    let rootEditUrl = `${projectHref}/edit/${primaryBranch}${projectPath}`;
+    let match = this._locateFile(path);
+    if (match) {
+      let { projectHref, addonPathInRepo, docsAppPathInRepo, primaryBranch } = config['ember-cli-addon-docs'];
```
ef4: Now instead of a single `projectPathInRepo`, we have separate `addonPathInRepo` and `docsAppPathInRepo`.

ef4:
As of https://github.com/ember-animation/ember-animated/pull/171 ember-animated is using this fork, so you can see how it works. It's deployed as https://ember-animation.github.io/ember-animated/versions/master/

samselikoff:
Amazing amount of effort + work. Thanks for donating so much to this community project!

Could you give us a tl;dr of the steps an Addon should/would take to move from the dummy app to a standalone `/docs` app?

ef4:
General steps to convert an existing dummy app to a separate app:

1. Decide whether you want a monorepo or a nested app. Either should work, but they have different pros and cons. With a monorepo you only need to run `yarn install` once and you get efficient sharing of dependencies. But it's also harder for users to point their app directly at an unreleased commit or fork of your addon. 

    In my case, I went with a nested app, meaning I have a regular ember-addon, plus a `/docs` folder that contains a regular ember app. So you need to `yarn install` in both before the docs app will work. 

2. Generate a new app to be your docs app, with `ember new`.  ember-cli doesn't like to generate new apps inside existing apps, so you may need to generate it elsewhere first and then move it into place.

3. Make your docs app depend on your addon. In my case, I did it as an [in repo addon](https://github.com/ember-animation/ember-animated/blob/152943adb41cdf2de8082b679daafffd6f9c3f77/docs/package.json#L70). If you're using a monorepo with yarn workspaces, you can have a regular `devDependency` on it instead.

4. Add `ember-cli-addon-docs` to your docs app.

5. Configure the docs app's `ember-cli-build.js` to tell ember-cli-addon-docs to point at the addon:

    ```js
    let app = new EmberApp(defaults, {
      'ember-cli-addon-docs': {
        documentingAddonAt: '..'
      }
    });
    ```

    At this point you should be able to run the docs app and it should look the same as if you had just added ember-cli-addon-docs for the first time to your addon. 

6. Move all the things (routes, templates, components, styles, etc) in your tests/dummy app into your new docs app. Also move any tests that are specifically about testing the docs app, as opposed to testing the addon. When you're done, you'll still have a dummy app and you can still use it for dev and test, but it won't contain the docs.

7. Remember to configure CI to run `ember test` in both the addon and the docs app. 

8. If you have a nested docs app, remember to configure `.npmignore` or equivalent so you don't publish the whole docs app to NPM.

ef4:
I think dependabot lets you configure individual directories in a repo, but somebody should try it.

rwwagner90:
@samselikoff are we good to merge this?

samselikoff:
I think it would be fine to merge but also would be good to get one other app testing it? Ideally even a test-project within this repo but if that's too much work just convert some other Addon to use a separate project.

josemarluedke:
I started working on a new set of addons and created an app for the docs. I tested this branch with my setup and it is working as expected. It would be awesome to get this merged into master and released.

My setup is here: https://github.com/josemarluedke/frontile/tree/docs-app

rwwagner90:
@samselikoff is that good enough to merge?

samselikoff:
Seems good! Will try to do this on one of my addons soon.

PR #414, opened 2019-10-19
Upgrade Project Dependencies

josemarluedke:
Most of the dependencies got updated to latest. There are a few that have breaking changes that should be done in another PR.

They are:

```
ember-code-snippet   ❯  2.4.2   ❯  3.0.0
ember-href-to   ❯  1.15.1  ❯  3.1.0
tailwindcss   ❯  0.7.4   ❯  1.1.2
eslint-plugin-ember   ❯  6.10.1  ❯  7.1.0
```


josemarluedke:
cc @dfreeman @samselikoff @@rwwagner90.

package.json
```diff
@@ -25,114 +25,112 @@
...
+    "ember-cli-sass": "10.0.1",
+    "ember-cli-string-helpers": "^4.0.5",
     "ember-cli-string-utils": "^1.1.0",
     "ember-code-snippet": "^2.4.1",
     "ember-component-css": "^0.7.4",
     "ember-composable-helpers": "^2.3.1",
     "ember-concurrency": "^0.9.0 || ^0.10.0 || ^1.0.0",
-    "ember-data": "2.x - 3.x",
-    "ember-fetch": "^6.7.0",
+    "ember-data": "^3.13.1",
```
rwwagner90: Do we want to enforce 3.13.1+ for ember-data? Since these are dependencies, and not devDependencies, I think keeping the range might be nice, so theoretically, it'll use whatever version you may already have. Not really sure if it makes a huge difference though. @samselikoff thoughts?

josemarluedke: That's fair. I will update.

rwwagner90: I think we should also probably keep the deps at whatever the blueprints set them to for addons, so we do not enforce a higher version unnecessarily, but I am not sure if it's worth worrying about. Just trying to keep things compatible with a wide range of versions out of the box.

samselikoff: Yep more liberal ranges are probably good, though I think https://github.com/ember-learn/ember-cli-addon-docs/pull/409 is the higher yield work to address this sorta thing.

package.json
```diff
@@ -25,114 +25,112 @@
...
     "broccoli-caching-writer": "^3.0.3",
     "broccoli-debug": "^0.6.4",
     "broccoli-filter": "^1.2.4",
     "broccoli-funnel": "^2.0.2",
     "broccoli-merge-trees": "^3.0.1",
     "broccoli-persistent-filter": "^2.3.1",
-    "broccoli-plugin": "^1.3.1",
-    "broccoli-source": "^1.1.0",
-    "broccoli-stew": "^2.0.0",
+    "broccoli-plugin": "^3.0.0",
```
dcyriller: May I suggest a change here as an addon user?

```suggestion
    "broccoli-plugin": "1.1.0 - 3",
```

If you specify a more permissive range in the dependencies, the consuming apps will be able to use either v1, v2 or v3. This will allow their dependencies to be deduplicated. This would be especially useful for those who use something like ember-dependency-lint (that makes your CI fail when your app has multiple versions of a same dependency).

This remark doesn’t concern the devDependencies as they are neither downloaded nor loaded by the consuming apps.

josemarluedke: Thanks for the catch. Just pushed a commit to revert to the original code here.

Issue #406, opened 2019-10-02
Addon clobbers tailwind configuration


alexlafroscia:
My addon is using `ember-cli-postcss` to manually configure Tailwind as a PostCSS plugin.

The configuration of that instance of Tailwind seems to be clobbering the Tailwind configuration provided by `ember-cli-addon-docs`. For example, I do not want my Tailwind configuration to include the CSS reset; if I configure the `corePlugins` to remove the `preflight` plugin, the `ember-cli-addon-docs` side does not render correctly as it _also_ has the reset removed.

Is there some way to allow the `ember-cli-addon-docs` styles not to be affected by the instance of `tailwind` that is being run on my own CSS?

alexlafroscia:
I’m guessing the reason is

- `addon-docs` adds a Tailwind plugin
- Addon adds it’s own Tailwind plugin
- PostCSS does not allow multiple versions of a plugin and inadvertently replaces the first with the second

I’ve looked around online and haven’t had luck finding information on how to tell PostCSS to run both. I went into the debugger and overrode the `postcssName` (or something) of one of the plugins but that didn’t fix the issue.

rwwagner90:
@alexlafroscia I think we need to do some work to use tailwind manually here, rather than ember-cli-tailwind, as per the discussion here https://github.com/embermap/ember-cli-tailwind/issues/143

I am not sure if that helps or not, but would be a step in the right direction.

alexlafroscia:
Would making that change be enough to get this working?

I'm happy to take a pass at dropping `ember-cli-tailwind` in this add-on if it can resolve my issue, as it's a bit of a blocker for us right now.

rwwagner90:
@alexlafroscia I am not sure if that would fix it, but I suspect it would be helpful. If you have the time to give it a try, that would be great!

alexlafroscia:
@rwwagner90 you confirmed that this isn't an issue anymore?

rwwagner90:
@alexlafroscia I did not, I thought I saw you commented somewhere that it was fixed?

alexlafroscia:
I believe it's fixed!

rwwagner90:
Closing for now, if anyone sees this again, please let us know!

PR #415, opened 2019-10-21
Remove sass in favor or plain css with PostCSS

josemarluedke:
This simplifies our styles and potentially will allow us to publish a single CSS file with all the compiled style, including Tailwind CSS. That should also help with #406. Although, I don't know how we would handle ember-component-css.


rwwagner90:
@josemarluedke I'm not sure I understand why this is preferred to sass?

alexlafroscia:
Tailwind recommends _not_ using Sass, if that matters at all.

I'm very interested in anything that avoids the `ember-cli-addon-docs`-provided version of Tailwind from being clobbered by an addon's own installation of it, but is that actually done in this PR?

> potentially will allow us to publish a single CSS file

It's unclear whether this was actually done in this change.

PR #416, opened 2019-10-22
dependabot: Add a config

dcyriller:
#### Description

Would you accept a Dependabot config to automate dependencies upgrade?

#### Details

This config runs dependabot **on a weekly basis** (not to be too verbose). It can also be run on a montly basis if you prefer.

This config should also **auto-merge in-range updates**. To enable that feature, you'll have to visit https://app.dependabot.com/accounts/ember-cli-addon-docs/settings and tick the box `Allow auto-merging to be enabled on projects`.

To enable **squash and merge** for dependabot PRs, you'll have to untick the box `Create a merge commit if a PR is merged by Dependabot`.

With the `version_requirement_updates: auto` option, Dependabot should:
- increase versions in the `devDependencies`
- widen ranges in the `dependencies`

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,11 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "weekly"
+    default_labels:
+      - "dependencies"
+    automerged_updates:
+      - match:
+          update_type: "in_range"
+    version_requirement_updates: "auto"
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -2,10 +2,13 @@ version: 1
 update_configs:
   - package_manager: "javascript"
     directory: "/"
-    update_schedule: "weekly"
+    update_schedule: "daily"
     default_labels:
       - "dependencies"
     automerged_updates:
       - match:
           update_type: "in_range"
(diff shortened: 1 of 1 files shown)


rwwagner90:
This LGTM! @samselikoff any objections?

samselikoff:
Looks good, I don't see any reason not to do `daily`. That's what we do on our projects.

Also it might be worth adding

```
    ignored_updates:
      - match:
          dependency_name: "ember-cli"
```

so we can do that ourselves, because I believe ember-cli updates typically are best done with `ember-cli-update` to bring along any changes to blueprint files with them. Here's an example from one of our projects: https://github.com/embermap/ember-cli-fastboot-testing/blob/master/.dependabot/config.yml

rwwagner90:
@samselikoff I think it might be good to ignore all deps that might be updated by ember-cli-update, and handle the ember updates separately. Would love to create a bot that runs `ember-cli-update`, but haven't had the time yet.

samselikoff:
Yeah we used to so that but it’s actually better to let the other deps float, ember cli is really the only one needed to be done manually. It’s easier that way.

rwwagner90:
@samselikoff fair enough 👍 @dcyriller can you update per Sam's comments?

PR #418, opened 2019-10-25
Refactor route anchor jump system

### LINKED ISSUE #417: Tests fail in FastBoot when addon uses addon docs + ember-cli-fastboot-testing
This issue was recently introduced here:

https://github.com/ember-learn/ember-cli-addon-docs/blob/master/addon/initializers/route-anchor-jump.js#L9

In FastBoot, there is no `window.location`, so this line causes tests that are written in an addon that uses `ember-cli-addon-docs` and has tests that `ember-cli-fastboot-testing` to fail.

```
Error while processing route: content-paint location is not defined ReferenceError: location is not defined
    at Class.afterModel (/tmp/broccoli-7127ipwpCiFn887X/out-481-srihash_assets/assets/addon-tree-output/ember-cli-addon-docs/initializers/route-anchor-jump.js:15:1)
```

One solution I can think of is to just not run this reopen if using FastBoot, since it's related to interactive behavior.

rwjblue:
* avoid making reopen inside `initialize` (it should only be done once **ever** not once per application)
* avoid destructuring from `location` when it is not defined (e.g. FastBoot contexts)
* ensure that the super result is properly returned (previously we would drop any promises returned from afterModel 😱)

Fixes #417

### DIFF
--- a/addon/initializers/route-anchor-jump.js
+++ b/addon/initializers/route-anchor-jump.js
@@ -1,11 +1,9 @@
 import Route from '@ember/routing/route';
 import { scheduleOnce } from '@ember/runloop';
 
-export function initialize() {
-  Route.reopen({
-    afterModel() {
-      this._super(...arguments);
-
+Route.reopen({
+  afterModel() {
+    if (typeof location !== 'undefined') {
       const { hash } = location;
       if (hash && hash.length) {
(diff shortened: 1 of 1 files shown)


scalvert:
LGTM! Thanks, Rob.

PR #419, opened 2019-10-25
Change scheduleOnce to schedule

scalvert:
Using `scheduleOnce` with an anonymous function [doesn't actually schedule only once](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md), since a new function is created on every invocation. This PR just changes to use `schedule`, which I believe is fine in this case.

### DIFF
--- a/addon/initializers/route-anchor-jump.js
+++ b/addon/initializers/route-anchor-jump.js
@@ -1,12 +1,12 @@
 import Route from '@ember/routing/route';
-import { scheduleOnce } from '@ember/runloop';
+import { schedule } from '@ember/runloop';
 
 Route.reopen({
   afterModel() {
     if (typeof location !== 'undefined') {
       const { hash } = location;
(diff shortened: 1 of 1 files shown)


rwjblue:
I missed this in my last PR, sorry about that...

scalvert:
How dare you, @rwjblue...

:)

rwjblue:
Yes, you can use a stable function and use `scheduleOnce`, but you can't just add a name to a function passed directly to `scheduleOnce`. 

In other words, this works fine:

```js
function foo() {}

// ...snip...
afterModel() {
  scheduleOnce('afterRender', foo);
}
// ...snip...
```

But this does not:

```js
// ...snip...
afterModel() {
  scheduleOnce('afterRender', function foo() {});
}
// ...snip...
```

PR #426, opened 2019-11-08
docs-hero: Fix YUIDoc property name logo→prefix to match actual property name

lolmaus:
cat Salvador singing "You're a dream of my days"


rwwagner90:
@lolmaus mind rebasing this as well?

lolmaus:
Green now.

PR #428, opened 2019-11-29
rootURL: replace backslashes with forward slashes, fixes #427

### LINKED ISSUE #427: Error: Could not read config from meta tag with name "dummy/config/environment" on Windows
Hi!

When I deploy to Github Pages and open the result in the browser, I receive this error:

> Error: Could not read config from meta tag with name "dummy/config/environment"

This is because the meta tag contains invalid JSON. Here's the offending fragment:

    "rootURL":"/ember-cli-yadda-opinionated/versions\addon-docs/"

As you can see, a reverse slash is used, which is treated as an escape character.

This looks like a Windows issue, as I'm building on Windows.

Related: https://github.com/ember-learn/ember-cli-addon-docs/issues/385

lolmaus:
This might be not the right place to do the replacement, but it does the thing and unblocks my deploys.

### DIFF
--- a/lib/deploy/plugin.js
+++ b/lib/deploy/plugin.js
@@ -151,7 +151,7 @@ module.exports = class AddonDocsDeployPlugin {
 
   _updateIndexContents(context, stagingDirectory, appRoot, deployVersion) {
     let indexPath = `${stagingDirectory}/${appRoot}/index.html`;
-    let rootURL = [this._getRootURL(), appRoot].filter(Boolean).join('/');
+    let rootURL = [this._getRootURL(), appRoot].filter(Boolean).join('/').replace(/\\/g, '/');
     let addonDocsRootURL = rootURL === '' ? '/' : `/${rootURL}/`;
     let contents = fs.readFileSync(indexPath, 'utf-8');
     let encodedVersion = encodeURIComponent(JSON.stringify(deployVersion));


rwwagner90:
@lolmaus this seems to break the tests. Can you elaborate on why this is needed? Is it that windows generates paths with backslashes, then we need to convert them back? If would be great if we could get the existing stuff to always return the right paths, rather than manually changing them.

lolmaus:
> this seems to break the tests.

I don't think so. My branch is red because the master has been red when I branched off.

This seems to be the offending commit: https://travis-ci.org/ember-learn/ember-cli-addon-docs/builds/607226331 At least it's commit message sounds related to the assertion error message.

***

> Can you elaborate on why this is needed? Is it that windows generates paths with backslashes, then we need to convert them back?

Yes.

***

> If would be great if we could get the existing stuff to always return the right paths, rather than manually changing them.

Well, yes, I generally agree, as I said in the top message. But I don't know where exactly the issue happens, and debugging Node is tedious.

My solution seems to be adequate and robust.

rwwagner90:
@lolmaus can you rebase this please so we can see if the tests pass?

lolmaus:
@rwwagner90 Passing now!

rwwagner90:
Thanks for the PR!

PR #433, opened 2019-12-09
Use double quotes everywhere

ctjhoa:
This PR replace all single quotes to double quotes.

**Why?**
- It wasn't consistent across this project
- Ember app comes with double quotes by default
- You can't copy paste examples without introducing linting errors in a new App


rwwagner90:
@ctjhoa I think we need to also update some tests?

ctjhoa:
> @ctjhoa I think we need to also update some tests?

Yeah I saw this...

PR #434, opened 2019-12-09
Fix examples to use `docs-container` class instead of `container`

ctjhoa:


tests/dummy/app/pods/docs/quickstart/template.md
```diff
@@ -69,7 +69,7 @@ AddonDocs comes with some components to help you showcase your addon. Let's add
 {{#docs-snippet name="quickstart-marketing-index.hbs" title="tests/dummy/app/templates/index.hbs"}}
   {{docs-hero}}
 
-  <div class="container">
+  <div class="docs-container">
```
rwwagner90: I don't think this is a desired change. The `docs-` prefix is supposed to be an internal thing, not used by consumers. At least that is my understanding. What do you think @samselikoff?

samselikoff: Yep that's correct. The quickstart should probably have its own styles defined to emphasize the point that styling beyond what's provided by the components are the responsibility of the host addon.

Also, I intended to add a `<AddonDocs::Container />` as a public component for a container that matches AddonDocs' own breakpoints, which would make the quickstart better.

rwwagner90: Should we close this as won't fix then?

samselikoff: @ctjhoa Could you update this example, to use inline styling for the container as a partial improvement over the current code in the quickstart? (It's a little misleading having the `container` class there because that's not something provided by AddonDocs).

```diff
- <div class="container">
+ <div style="max-width: 40rem; margin: 0 auto; padding: 0 1.5rem">
```

ctjhoa: @samselikoff done

PR #435, opened 2019-12-12
Upgrade to tailwind 1.x

psbanka:
According to [this comment](https://github.com/ember-learn/ember-cli-addon-docs/issues/178#issuecomment-516514398), we need to upgrade ember-cli-addon-docs to tailwind v1.0  in order to take advantage of Tailwind's configuration directive which will allow us to turn off the ~~preflight~~ base styles by adding a configuration option of

```javascript
corePlugins: {
  container: false,
  preflight: false
}
```
to address the core issue found in https://github.com/ember-learn/ember-cli-addon-docs/issues/178#


samselikoff:
Nice work Peter!

I think this still doesn't allow users to disable base styles yet right? That will be another feature?

psbanka:
That's right @samselikoff — I'm working on a follow-up PR right now to make the inclusion of the presets configurable.

PR #436, opened 2019-12-13
Upgrade ember-code-snippet to v3.0.0

josemarluedke:
This was a major version bump with some breaking changes: https://github.com/ef4/ember-code-snippet/blob/master/CHANGELOG.md#300

The notable change here is that they don't supply syntax highlight component out of the box anymore. We had highlight.js already, so it was just a matter of wiring everything together. 

Note on the minimal style changes, I was looking at MirageJS docs site, and copied some changes from them as IMO looks a bit better.

cc/ @samselikoff @rwwagner90


rwwagner90:
@josemarluedke does this change the user facing API at all?

josemarluedke:
Nope, no changes for addon-docs consumers.

samselikoff:
Wow, thanks so much for doing this @josemarluedke ! Been meaning to work on this forever.

Looks like there are some failures in 2.12/2.18 🤔

josemarluedke:
@samselikoff tests are passing now. The let helper is not available in ember v2.x.

samselikoff:
Thanks man, nice work!

PR #438, opened 2019-12-17
WIP: Optional preflight styles

psbanka:
I could use some help on this PR because the current system *does not work*: 

- I used some broccoli build-tweaking in `ember-cli-build.js` to creates a `tailwind.config.js` file with the necessary configuration options set. **However**,  despite the fact that this file is built properly, that file does not seem influence whether or not preflight styles actually get included. 
- Something seems to be wrong with respect to when the postcss/tailwind files are run
versus when the `tailwind.config.js` file is processed.

Any anyone provide any advice for how to tackle this build problem or even suggest completely different approaches for how to configure the `tailwinds.config.js` at build-time?

### DIFF
--- a/addon/styles/tailwind.config.js
+++ b/addon/styles/tailwind.config.js
@@ -834,7 +834,9 @@ module.exports = {
   plugins: [],
 
   corePlugins: {
-    container: false
-  },
-
+    container: false,
+    // @ifdef NO_BASE_STYLES
+    preflight: false
+    // @endif
+  }
 };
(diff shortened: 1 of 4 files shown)


josemarluedke:
I haven't given much thought on this, but maybe using Embroider Macros could help here. 

Note that Embroider Macros works with traditional ember-cli, not required to have embroider.

https://github.com/embroider-build/embroider/tree/master/packages/macros

samselikoff:
I'm wondering if there's a different approach we could take, where AddonDocs can make the different parts of its styling available to the host addon to `app.import` as they please. That way they can just choose which pieces to include or not.

```js
app.import('ember-cli-addon-docs/styles/tailwind-base');
app.import('ember-cli-addon-docs/styles/tailwind-components');
app.import('ember-cli-addon-docs/styles/tailwind-utilities');
app.import('ember-cli-addon-docs/styles/component-styles');
```

PR #452, opened 2020-01-30
fix array helper clobbering built in helper

kturney:
The version of ember-composable-helpers currently being used provides an array helper.
ember@next has started erroring when it encounters built in helpers being clobbered.

ember-composable-helpers@3 has removed the array helper and upgraded to octane.
As an alternative to the commit, we could try to get a new version of ember-composable-helpers@2 cut that excludes the array helper for ember <= 3.8.

### DIFF
--- a/index.js
+++ b/index.js
@@ -163,6 +163,34 @@ module.exports = {
 
     includer.options.snippetExtensions = snippetExtensions;
 
+    const VersionChecker = require('ember-cli-version-checker');
+    const checker = new VersionChecker(this.project);
+    const ember = checker.forEmber();
+
+    if (ember.gte('3.8.0')) {
+      // array helper is built into ember as of 3.8.
+      // ember 3.17 starts erroring when overwritting built in helpers
+
+      // exclude from ember-cli-addon-docs
+      this.options['ember-composable-helpers'] = {
+        except: ['array']
+      };
+
+      // exclude from the app using ember-cli-addon-docs
+      if (!includer.options['ember-composable-helpers']) {
+        includer.options['ember-composable-helpers'] = {};
+      }
+
+      const echOptions = includer.options['ember-composable-helpers'];
+      if (!echOptions.only) {
+        if (!echOptions.except) {
+          echOptions.except = ['array'];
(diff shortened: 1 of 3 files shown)


kturney:
Additionally, I made an attempt at tracking down the timeout in test `component page index works`.
For some reason [the click on the 'Internal' toggle](https://github.com/ember-learn/ember-cli-addon-docs/blob/e98dfeed6dcaeb8ba13acf4145dd2d8750f33597/tests/acceptance/sandbox/api/components-test.js#L35) is causing both the internal toggle and the private toggle to be checked (but also weirdly doesn't trigger a `change` event on the private checkbox.).
So when [the 'private' toggle is clicked](https://github.com/ember-learn/ember-cli-addon-docs/blob/e98dfeed6dcaeb8ba13acf4145dd2d8750f33597/tests/acceptance/sandbox/api/components-test.js#L42) the checkbox is set to checked=false, so when [we update in x-component](https://github.com/ember-learn/ember-cli-addon-docs/blob/e98dfeed6dcaeb8ba13acf4145dd2d8750f33597/addon/components/api/x-component/component.js#L48) the value doesn't actually change (setPrivate already equals false and is set to false).
If an additional private toggle click is placed after the existing private toggle click, the test passes.

aklkv:
Any update on this!

rwwagner90:
@kturney can you please rebase this off of master? Need to see if we can get Travis to run.

kturney:
Hi @rwwagner90, I've rebased this PR on master.

I've also rebased https://github.com/ember-learn/ember-cli-addon-docs/pull/451 and https://github.com/ember-learn/ember-cli-addon-docs/pull/450 if you have any time to look at those.

rwwagner90:
@kturney thanks! Looks like we have the issue with Ember and Babel now. I know that's fixed in latest Ember releases, but might take a bit to get Travis totally fixed here.

ming-codes:
Can we cut a release for this?

rwwagner90:
@ming-codes not until we get a passing build of the addon.

PR #456, opened 2020-02-28
Fix doc typo

elwayman02:
`docs/template.hbs` is the incorrect file name, the file it asks you to create should match the title of the `docs-snippet`


elwayman02:
Yes, but pods are not the community-supported pattern. The official documentation should reflect the way that the community writes code according to the official Ember guides (classic and/or Octane)

elwayman02:
Also, it's internally inconsistent, as you can see the snippet tells you a different filepath altogether.

rwwagner90:
@elwayman02 mind rebasing with master please?

PR #470, opened 2020-04-30
Upgrade ember-composable-helpers

jacojoubert:
This fixes the error `You attempted to overwrite the built-in helper "array" which is not allowed`


rwwagner90:
@jacojoubert I believe this was already fixed here https://github.com/ember-learn/ember-cli-addon-docs/pull/452

jacojoubert:
Aah I did not see that. Strange as using the master version didn't seem to work. I will close.

rwwagner90:
@jacojoubert I never checked that the other PR truly fixed it, if more work is needed, we can definitely look at it.

PR #475, opened 2020-06-18
ember-cli 3.18, some lint fixes

rwwagner90:
I *think* the removal of some `this.get` might be classified as "breaking", since native getters dropped around 3.1, but I would propose that we only support the Ember versions in the latest addon blueprints at any given time.


samselikoff:
I don't have time to maintain AD anymore and don't want to go radio silent on you! I'm happy to give any input on direction or design questions if I can be of help but I won't be able to review PRs.

Just wanted to let you know so I don't block any of your work!

PR #552, opened 2020-09-07
Add ember-data as a peer dep

snewcomer:
https://github.com/embermap/ember-data-storefront/pull/396/checks?check_run_id=1079154106

Looking at a build error for what I assume is b/c ember-data is not listed as a peer dep (to allow the consuming library to specify).  Also, if it is a peer dep, does it require as a dependency as well (currently in the `dependencies` block as well)?

Another solution might be adding "ember-data" to a resolutions block in the repo above.  

Let me know what you think!


rwwagner90:
@snewcomer I think if we make it a peerDep, that we would want it as a devDep of just the latest version, and remove from deps.

rwwagner90:
@snewcomer I wonder if we should take this opportunity to also only support ember-data `>= 3`? I'm working on making this addon be less broad on support, in favor of a more simplified addon.

snewcomer:
Certainly!  Do you see this going in a minor release or major?  If the former, then we can merge and I can follow up!

rwwagner90:
@snewcomer I think restricting to 3+ would mean we need a major right?

snewcomer:
I think so!  I can make a follow up PR then if you think that is the right path forward.

rwwagner90 (approved):

rwwagner90:
@snewcomer sure thing, will merge this one for now. Please do feel free to follow up.

PR #553, opened 2020-09-07
Minimum 3.x series for ember-data addon

snewcomer:
ref #552 

I am unsure if peer dependencies is a strict versio or loose range.  Might be worthwhile to make it `3.x - and above` if so.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
     "release-it-lerna-changelog": "^2.3.0"
   },
   "peerDependencies": {
-    "ember-data": "3.x",
+    "ember-data": ">= 3.0.0",
     "ember-fetch": "^8.0.2"
   },
   "engines": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -143,7 +143,7 @@
     "release-it-lerna-changelog": "^2.3.0"
   },
   "peerDependencies": {
-    "ember-data": "2.x - 3.x",
+    "ember-data": "3.x",
```
rwwagner90, line 146: I think we should do `>= 3.0.0`, but not sure. I would want it to work with 4.x too later potentially.

rwwagner90 (approved):

snewcomer:
Great! Thx for the review.  Possible we can get a 0.10 release before this goes in?

rwwagner90:
@snewcomer sure, I can do another release before this.

PR #585, opened 2020-09-29
Add test for both ember-concurrency v1 and v2

maxfierke:
We've just released ember-concurrency 2.0.0-beta.1. It's a largely API-compatible re-engineering of ember-concurrency's internals to support tracked properties and decouple from Ember (shedding EmberObject, etc.). As a popular library which depends on ember-concurrency, it would be nice for `ember-cli-addon-docs` to support both 1.x and 2.x versions concurrently so that folks can upgrade the version of ember-concurrency in their applications at their leisure without waiting for all the add-ons they use to switch over to a newer version (and risk leaving folks who cannot upgrade yet behind.)

This PR adds some scenarios to ember-try to enable testing against both major releases and relaxes the version restriction.

The build fails for beta and canary, but does not appear related to these changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,8 @@ jobs:
     - env: EMBER_TRY_SCENARIO=ember-canary
     - env: EMBER_TRY_SCENARIO=ember-default-with-jquery
     - env: EMBER_TRY_SCENARIO=ember-classic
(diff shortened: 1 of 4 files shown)


rwwagner90 (approved):

rwwagner90:
Thanks for the PR!

PR #626, opened 2020-11-03
Migrated CI/CD from Travis to GitHub Actions

ijlee2:
## Description

Over the past couple of weeks or so, we noticed that the queue for Travis CI runs has been backing up. For `ember-cli-addon-docs`, Travis CI runs would take about 60 minutes. `dependabot` has been opening many pull requests recently; I think this may have caused the queue to back up for other repos.

In this pull request, we consider migrating CI to GitHub Actions.

## Notes

I met feature parity with the current Travis CI configuration. (I didn't replicate "floating dependencies" tests, however.)

The new workflow can do the following in parallel:

- Run `lint:hbs`, `lint:js`
- Run `test:ember`, `test:node`, `test:test-apps`
- Run `test:compatibility` (`ember-try` scenarios)
- Cache installed packages
- Deploy app (documentation page) when a pull request is successfully merged to `master` branch

## How to Review

I recommend reviewing the commits one by one. Each commit makes a small, meaningful change.

## References

- [Workflow template for Ember apps](https://gist.github.com/ijlee2/1c864ebe96a55f239e80800829ef0bf4)
- [Workflow template for Ember addons](https://gist.github.com/ijlee2/25004d700cd008ee08ecc4873c109a5b)
- https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing


package.json
```diff
@@ -114,7 +114,6 @@
     "ember-cli-deploy": "^1.0.2",
     "ember-cli-deploy-build": "^2.0.0",
     "ember-cli-deploy-git": "^1.3.4",
-    "ember-cli-deploy-git-ci": "^1.0.1",
```
ijlee2, line 117: With GitHub Actions, we don't need `ember-cli-deploy-git-ci`. For more information, please see sections 3b and 3c of https://crunchingnumbers.live/2020/08/31/ci-with-github-actions-for-ember-apps-part-2/.

package.json
```diff
@@ -15,15 +15,15 @@
...
     "lint": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*",
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "prepare": "./scripts/link-them.sh",
     "start": "ember serve",
     "test": "npm-run-all lint:* test:*",
-    "test:browser": "ember test --test-port=0",
-    "test:ember": "ember test",
-    "test:ember-compatibility": "ember try:each",
+    "test:ember": "ember test --test-port=0",
```
ijlee2, line 25: I'm not sure why previously `--test-port=0` was needed. Both `ember test --test-port=0` and `ember test` ran fine locally. We may be able to write `ember test` to more closely follow the `package.json` in a new addon.

.github/workflows/ci-cd.yml
```diff
@@ -1,6 +1,6 @@
 name: CI/CD
 
-on: 
+on:
```
ijlee2, line 3: I made the 5th commit to check whether the new caching step in the 4th commit worked (fixed the error found in the 3rd commit).

jenweber (approved):
I haven't been involved in maintaining this repo, so this should be reviewed by Robert before merging, but here's my 👍 on migrating to GitHub actions.

ijlee2:
Thanks! Yep, I agree with getting a review from Robert.

.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,187 @@
...
+      - name: Use Node.js ${{ env.NODE_VERSION }}
+        uses: actions/setup-node@v2-beta
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Get Yarn cache path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - name: Cache Yarn cache and node_modules
```
rwwagner90, line 73: @ijlee2 I think perhaps we don't want to cache everything every time? I think we previously had some tests for locked and floating deps. Also, it seems most steps do the same steps of checkout, use node, install etc. Can we eliminate some of these installs and use the already installed stuff?

rwwagner90 (changes requested):

.github/workflows/ci-cd.yml
ijlee2: I appreciate your feedback! Let me address them in reverse order:

- To my understanding, currently it's not possible to do a "before each" hook (YAML anchor) in GitHub Actions. Having duplicate steps across different jobs is unfortunate but necessary.

    A workaround may be possible with a [composite run steps action](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/about-actions#composite-run-steps-actions), but I think it's better to duplicate code (simple approach) than abstract by creating an action (more complex). (There may be an action within the Ember community and I haven't found it yet.)

- I didn't understand why floating dependencies tests were needed for an addon. From the [ember-try documentation](https://github.com/ember-cli/ember-try#a-note-on-npm-scenarios-with-lockfiles), it does seem like we do want to continue to have floating dependencies tests. I can work on updating the workflow to restore this feature.

- For locked dependencies tests (if I understood what locked dependencies tests do versus floating), I think caching makes sense and will help decrease the runtime.

Can you help me understand the locked vs. floating tests better?

rwwagner90: I do not know personally why locked vs floating matters, but I think it's just to make sure the latest versions of deps are not breaking things, so you can catch it early.

I agree that caching while using locked deps is good 👍 

For skipping some duplicate steps, I think it is possible to use output from a previous step in a subsequent step. I.e. we should be able to choose the node version once, instead of every time, but I do not know the specifics myself.

ijlee2:
Travis' documentation (in particular, [Listing Individual Jobs](https://docs.travis-ci.com/user/build-matrix/#listing-individual-jobs) and [Build Stages](https://docs.travis-ci.com/user/build-stages/)) helped me understand what the current build does.

I believe I added 3 extra jobs for `test:ember`, `test:node`, and `test:test-apps` with floating dependencies. The screenshot below shows that the log messages for installation are different when dependencies are floating (left) versus locked (right). Maybe this is proof? 🤔

comparison of package installation with floating and locked dependencies

ijlee2:
@rwwagner90 Can you have a look at the latest commit, which addresses adding floating dependencies tests back?

I'd like to get the pull request merged soon so that there won't be merge conflict in `yarn.lock` when dependabot opens new pull requests.

PR #693, opened 2020-12-23
Remove pods support

rwwagner90:
Instead of supporting pods, which have been recommended against almost since they came into existence, we should support just the documented and supported formats Ember provides. This will give us less to maintain, and hopefully the more we scope down, the easier it will be to support the latest and greatest things.


knownasilya:
😢 Mainly because there is no good alternative to pods.

rwwagner90:
@knownasilya apologies, but since pods have been recommended against for years, I felt the overhead of supporting them was not worth it and was more confusing for users getting started. I personally like component colocation and keeping everything else classic, but I know a lot of folks do like pods still, so sorry for this 😕

PR #941, opened 2021-09-17
build: remove some unused deps / devDeps

ndekeister-us:
### Remove unused dependencies

- [broccoli-debug](https://github.com/broccolijs/broccoli-debug)
    - Added in https://github.com/ember-learn/ember-cli-addon-docs/pull/247 via this commit https://github.com/ember-learn/ember-cli-addon-docs/pull/247/commits/9de2b590407c0b21cf21d40241716fd63996e8bf
    - Not used since the same PR https://github.com/ember-learn/ember-cli-addon-docs/pull/247

- [inflected](https://github.com/martinandert/inflected)
    - Added in https://github.com/ember-learn/ember-cli-addon-docs/commit/2d0e288f9cb223f7fbe680aeaf4f37d96250ef58#diff-48394617c0240fb029fa5459bf8a84c5a7a5e64588344681183a41bf60e647fbR9
    - Not used since https://github.com/ember-learn/ember-cli-addon-docs/commit/0561c3d9d82946b6f0c68adf646010c494d3488a#diff-d34fb7527260048f2ddde586f0d54dbce0ccb6dda8093c96dcedaa8ea8688a8cL8

- [resolve](https://github.com/browserify/resolve)
    - Added in https://github.com/ember-learn/ember-cli-addon-docs/commit/769b219413967804807fecbd726ae656fbb7c25c
    - Not used since https://github.com/ember-learn/ember-cli-addon-docs/commit/fac74c03c67d195bbc15f1e21c7bc894ae4b3695

### Remove unused devDeps

- [ember-page-title](https://github.com/ember-cli/ember-page-title)
    - Not used in the dummy app, we could also update the dummy app to add correct title on each page?

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,6 @@
     "ember-export-application-global": "^2.0.1",
     "ember-load-initializers": "^2.1.2",
     "ember-maybe-import-regenerator-for-testing": "^1.0.0",
-    "ember-page-title": "^6.2.2",
     "ember-qunit": "^5.1.4",
     "ember-resolver": "^8.0.3",
     "ember-source": "~3.28.0",
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "highlight.js": "^10.7.2",
     "hosted-git-info": "^4.0.2",
     "html-entities": "^2.3.2",
-    "inflected": "^2.0.3",
     "jsdom": "^16.4.0",
     "json-api-serializer": "^2.2.1",
     "lodash": "^4.17.15",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "@glimmer/syntax": "^0.80.0",
     "broccoli-bridge": "^1.0.0",
     "broccoli-caching-writer": "^3.0.3",
-    "broccoli-debug": "^0.6.5",
     "broccoli-filter": "^1.3.0",
     "broccoli-funnel": "^3.0.8",
     "broccoli-merge-trees": "^4.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -131,7 +127,6 @@
     "ember-export-application-global": "^2.0.1",
     "ember-load-initializers": "^2.1.2",
     "ember-maybe-import-regenerator-for-testing": "^1.0.0",
-    "ember-page-title": "^6.2.2",
```
ndekeister-us, line 134: We are not using it yet in the Dummy app, but maybe we should refactor the dummy app to add correct title for each page instead of removing this addon?

rwwagner90: I agree with removing it, we don't need it.

package.json
```diff
@@ -108,7 +105,6 @@
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.4.2",
     "@embroider/test-setup": "^0.44.1",
-    "@fullhuman/postcss-purgecss": "^4.0.3",
```
rwwagner90, line 111: We want to keep this for tailwind, I think. If it is not currently working, we should work on making sure we get purging working with tailwind

rwwagner90 (changes requested):

package.json
ndekeister-us: +1 made some tests locally and purging don't work at the moment, i reverted the commit in this PR and will try to tackle this "issue" in another PR, have something "basic" working where the extra CSS is purged for this repo and for repo using this addon but it will for sure require more testing and some refactor (like purging only when we build in production)

rwwagner90 (approved):

PR #950, opened 2021-09-27
fix: scss purging

ndekeister-us:
Fix scss purging, enabling it on production build only for both ember-cli-addon-docs and project consuming this addon ; also moved https://github.com/FullHuman/purgecss/tree/main/packages/postcss-purgecss to dependencies as it is now used in the `index.js`

On [ember-cli-addon-docs](https://github.com/ember-learn/ember-cli-addon-docs) (left before, right after)
purcecss-ember-cli-addon-docs

On [ember-exam](https://github.com/ember-cli/ember-exam/) (left before, right after)
purgecss-ember-exam

On [hex](https://github.com/mupkoo/hex) (a UI framework, docs made with Ember ; left before, right after)
purgecss-hex

___

Tested locally enabling it in dev. environment, saw no regression on both this addon and `ember-exam` docs (edit and no regression anymore on `hex`)

___

Todo
- check on one or two other projects using this addon
  - [x] test on https://github.com/mupkoo/hex
    - ~~some issue with `docs-md__1` classes deleted from vendors.css. If i understand well we automatically convert `# title` to `<h1 class="docs-md__1">`, so `docs-md__1` is not explicitly present in some js/hbs files and purgescss can't find it, so it delete it. I believe that we could add `/*! purgecss start ignore */` to `styles/tailwind/components/_docs-md.scss` file which will solve the issue, but need to better understand the behavior~~ -> in the end added `/*! purgecss start ignore */` to all our custom styles, to prevent similar / unwanted issues (in all case it don't have an huge impact on output file size to keep these custom styles)

### DIFF
--- a/index.js
+++ b/index.js
@@ -251,6 +251,7 @@ module.exports = {
               : []),
           ],
           css: [styleDir],
+          safelist: ['ember-modal-dialog'],
           defaultExtractor: (content) =>
             content.match(/[A-Za-z0-9-_:/]+/g) || [],
         },
--- a/index.js
+++ b/index.js
@@ -34,27 +34,6 @@ module.exports = {
           require('tailwindcss')(
             path.join(__dirname, 'addon', 'styles', 'tailwind.config.js')
           ),
-          {
-            module: require('@fullhuman/postcss-purgecss'), // eslint-disable-line node/no-unpublished-require
-            options: {
-              content: [
-                path.join(
-                  __dirname,
-                  '{addon,app,tests/dummy/app}',
-                  '{**/,}*.{js,hbs,html,md,ts}'
-                ),
-                path.join(
-                  __dirname,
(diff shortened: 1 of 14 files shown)


index.js
```diff
@@ -226,6 +227,23 @@ module.exports = {
     importer.import('vendor/lunr/lunr.js', {
       using: [{ transformation: 'amd', as: 'lunr' }],
     });
+
+    if (EmberApp.env() === 'production') {
```
ndekeister-us, line 231: Should we add some tests to ensure the behavior on production/development environment?

ndekeister-us, line 231: Should we add some tests to ensure the behavior on production/development environment?

rwwagner90:
@ndekeister-us thanks for this! Looking at the screenshots, it seems the CSS got larger in the first one?

PR #978, opened 2021-10-20
Fixes broken deployment with ember-auto-import@2

scalvert:
When using ember-auto-import@2, it will use the `rootURL` to prefix the URLs of the injected chunks. It does this without adding a preceding slash, which means the URLs are invalid and chunks cannot be loaded. As a result, required dependencies, such as highlight.js, are not available (since they weren't loaded by the preceding chunk).


rwwagner90 (approved):

miguelcobain:
Can this be released, please? 🙏

rwwagner90:
@miguelcobain unfortunately, we have a broken build right now, but hopefully it can be fixed soon.

PR #1006, opened 2021-11-20
chore(deps): drop ember-href-to

gilest:
Reason being that it imports `@ember/routing/link-component` which has been removed in ember v4.

Replaces the helper function from this addon with a direct call to the routing service.


rwwagner90:
Thanks @gilest! It looks like there are still some failures for beta and canary. I'm unsure if these are related or not. I need to do some digging.

rwwagner90:
I'm going to go ahead and merge this, in the interest of making progress. We can debug the beta and canary failures after it is merged. Thanks for the PR!

PR #1021, opened 2021-11-30
Fix LinkTo usage and explicitly provide `@route` and `@models` arguments

SergeAstapov:


addon/components/api/x-section/index.js
```diff
@@ -2,16 +2,14 @@ import Component from '@ember/component';
...
   @class Api/XSection
   @hide
 */
-export default Component.extend({
-  tagName: '',
+export default class ApiXSection extends Component {
+  tagName = '';
 
-  showImportPaths,
+  showImportPaths = config['ember-cli-addon-docs'];
```
rwwagner90, line 12: This does not seem correct. `showImportPaths` was destructured before.

addon/components/docs-header/search-result/index.hbs
```diff
@@ -1,12 +1,12 @@
-{{! template-lint-disable no-unknown-arguments-for-builtin-components }}
 <div {{on 'click' this.onClick}} {{on 'mouseenter' this.onMouseEnter}} data-test-search-result>
   <LinkTo
     class={{
       concat
       'docs-block docs-py-2 docs-px-3 docs-text-black docs-no-underline hover:docs-bg-grey-lighter '
       (if @selected 'docs-bg-grey-lighter')
     }}
-    @params={{this.linkArgs}}
+    @route={{this.linkArgs.route}}
```
rwwagner90, line 8: I think I had split these args out like this before and it broke things. Are you sure this works?

addon/components/docs-viewer/x-main/index.hbs
```diff
@@ -41,7 +42,8 @@
       <LinkTo
         class='docs-text-grey-darkest docs-text-large-4 docs-font-light docs-no-underline
         docs-border-b docs-border-grey hover:docs-border-grey-darkest docs-transition'
-        @params={{this.docsRoutes.next.route}}
+        @route={{this.docsRoutes.next.route}}
```
rwwagner90, line 45: Same with these

rwwagner90 (changes requested):

addon/components/api/x-section/index.js
SergeAstapov: @rwwagner90 ah yes, good catch!
Actually I'll revert this change, I planned to make native class conversion in separate PR to make things focused

SergeAstapov: reverted!

addon/components/docs-header/search-result/index.hbs
SergeAstapov: @rwwagner90 yes, this works in combination with `@models` argument

addon/components/docs-viewer/x-main/index.hbs
SergeAstapov: @rwwagner90 yes, this works in combination with `@models` argument

PR #1010, opened 2021-11-22
Update linting, bump deps

rwwagner90:

PR #1031, opened 2021-12-02
Add babel.config.js to .npmignore

SergeAstapov:
Follow up to #1010.

I may be wrong, followed https://github.com/babel/ember-cli-babel/issues/418, https://github.com/ember-modifier/ember-modifier/issues/102 and https://github.com/ember-modifier/ember-modifier/pull/101/files

My assumption is that addons should not ship `babel.config.js` (at least yet).

Also followed what Chris did with `buildEmberPlugins()` instead of specifying only `@babel/plugin-proposal-decorators` plugin, I assume it just gives us some sort of safety net and be on pair with ember-cli-babel (at least from babel config perspective)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -57,6 +57,7 @@ module.exports = {
         './lib/**/*.js',
         './tests/dummy/config/**/*.js',
         './tests-node/**/*.js',
+        'babel.config.js',
       ],
       parserOptions: {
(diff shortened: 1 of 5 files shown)


rwwagner90 (approved):

rwwagner90:
Thanks for this @SergeAstapov!

PR #1034, opened 2021-12-03
Update CHANGELOG.md with v4.2.1

SergeAstapov:
Looks like `lerna-changelog` does not pick up PRs with `dependency` label


rwwagner90:
We don't manually edit the changelog. If you want PRs to show up in the changelog, the labels must be what you expect in the changelog. If you want this to show as an enhancement, you can update the label on the PR and I think it should fix the changelog next time we run it.

SergeAstapov:
@rwwagner90 aha, makes sense. Thank you!

PR #1108, opened 2022-01-29
Fix no implicit this deprecation in docs-demo component

nickschot:
There was a missing `this` in the DocsDemo component making it fail to compile with ember-source v4+.

### DIFF
--- a/addon/components/docs-demo/index.hbs
+++ b/addon/components/docs-demo/index.hbs
@@ -1,4 +1,4 @@
-<div class="docs-rounded docs-border docs-border-grey-lighter docs-my-8 {{class}}">
+<div class="docs-rounded docs-border docs-border-grey-lighter docs-my-8 {{this.class}}">
 
   {{yield (hash
     example=(component "docs-demo/x-example")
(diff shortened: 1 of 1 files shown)


rwwagner90:
@nickschot I think this should be `@class`, but I'm not 100% sure. I don't see `class` defined in the JS for the component though.

nickschot:
Both works for classic components 😅 I'm fine with either here.

addon/components/docs-demo/index.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="docs-rounded docs-border docs-border-grey-lighter docs-my-8 {{class}}">
+<div class="docs-rounded docs-border docs-border-grey-lighter docs-my-8 {{this.class}}">
```
rwwagner90, line 1:
```suggestion
<div class="docs-rounded docs-border docs-border-grey-lighter docs-my-8 {{@class}}">
```

miguelcobain:
Would be nice to see this released for all of those wanting to use addon-docs with ember 4+. :D

rwwagner90:
@miguelcobain hmm, I am pretty sure I have passing tests for all my addons using addon docs, so it's odd that I am not seeing this. Happy to do a release though

rwwagner90:
4.2.2 has been released

miguelcobain:
Well, the tests pass with ember 4, that's true.
I was referring to running the dummy app with ember 4. I got runtime errors because of this (I believe this was the cause).

nickschot:
Yep! Same. There's more stuff to update, but as far as I could see this was the only really breaking thing for running the dummy app.

rwwagner90:
@miguelcobain @nickschot I would love help updating the docs to be angle bracket and anything else you might see that needs updating, if you have the time ❤️

PR #1117, opened 2022-02-08
update controllers syntax

miguelcobain:
Not much here, but trying to update the various addon parts one at a time.
I removed the `addon/controllers/docs/api/class.js` controller, since I didn't see any usage of it.


SergeAstapov:
`test-apps/new-addon/tests/dummy/app/controllers/route-using-compile-markdown.js` looks good!

would leave call about `docs/api/class` controller to @rwwagner90

rwwagner90:
@miguelcobain I would think that controller would be used somewhere, but I don't know. Looks like to me it's the controller to automatically generated method names for the auto generated docs?

miguelcobain:
@rwwagner90 a global search for `methodParams` yields no results in the whole repository.
Would that controller be used on the consuming app?

rwwagner90:
@miguelcobain perhaps you are right. It could be a leftover thing.

miguelcobain:
@rwwagner90 tried searching on ember observer for `methodParams`. No results that seem to use this. Same for `docs/api/class`, so nothing is importing the controller.

rwwagner90 (approved):

PR #1125, opened 2022-02-09
update docs-logo component

miguelcobain:


addon/components/docs-logo/index.js
```diff
@@ -35,8 +35,8 @@ export default class DocsLogo extends Component {
   @equal('logo', 'ember-data')
   showEmberData;
 
-  didReceiveAttrs() {
-    super.didReceiveAttrs(...arguments);
+  constructor() {
```
miguelcobain, line 38: @rwwagner90 Just realized that this won't throw if the user changes the argument to an invalid logo. Only throws on first render.

I think the chances of someone doing that are pretty slim though. But if you think it's worth it, I can have this assertion on a `did-update` render modifier.

rwwagner90 (approved):

addon/components/docs-logo/index.js
rwwagner90: @miguelcobain no, I think just throwing initially is fine

PR #1128, opened 2022-02-10
update docs header

miguelcobain:
Updates `<DocsHeader/>` component and its sub-components.


addon/components/docs-header/search-result/index.js
```diff
@@ -1,127 +1,114 @@
...
-              if (keyword.toLowerCase().indexOf(normalizedTerm) !== -1) {
-                let index = normalizedKeyword.indexOf(normalizedQuery);
-                matches.push(
-                  this._highlight(keyword, index, normalizedQuery.length)
-                );
-              }
-            });
-          }
-        });
+            matches.push(matchingText);
```
miguelcobain, line 65: There is a bug somewhere around this logic of finding matches for the searched text.
It is however not related to this update, since the problem is already present on ec-addon-docs.

Basically, it presents very weird matches for the query you type. I suggest we address this separately.

addon/styles/addon.scss
```diff
@@ -57,7 +57,6 @@ body > .ember-view {
   @apply docs-flex;
   @apply docs-flex-col;
 
-  @apply docs-mt-1;
```
miguelcobain, line 60: When you open the version selector, you could see the page jumping down a little bit (you can see that happening right now on the current docs website).

This was because opening the dropdown adds a new `.ember-view` div on the body, which gets targeted by this selector. Deleting the top margin didn't result in anything weird for me and fixed the problem.

SergeAstapov (approved):

rwwagner90 (approved):

PR #1130, opened 2022-02-10
update docs-link

miguelcobain:
Migrates the `<DocsLink/>` component to a glimmer component. This makes it possible to not depend on `@ember/legacy-built-in-components`.

This also makes this component "officially" **public** by adding documentation, making it show on the docs site.
Please see inline comments.


addon/components/docs-link/index.hbs
```diff
@@ -0,0 +1,27 @@
...
+      this.isRouteAndModelOnly
+      (component "link-to" route=@route model=@model)
+      (if
+        this.isRouteAndModelsOnly
+        (component "link-to" route=@route models=@models)
+        (component "link-to")
+      )
+    )
+  )
+  as |DocsLinkTo|
```
miguelcobain, line 15: You're probably wondering why the hell this is needed.

My first try was
```hbs
<LinkTo
  @route={{@route}}
  @model={{@model}}
  @models={{@models}}
  @query={{@query}}
  @disabled={{@disabled}}
  @activeClass={{@activeClass}}
  @current-when={{@current-when}}
  ...attributes
  >
  {{yield}}
</LinkTo>
```

basically proxying every argument that `<LinkTo>` supports. However, this doesn't work because `<LinkTo>` is very picky about the arguments we pass to it.

First, we can't pass in `@model` and `@models` at the same time, even if one of them is `undefined`.
Then, even if you just pass `@route="something" @model={{undefined}}`, the link also won't work. You really need to **only** pass in `@route="something"` for those cases.

So, what I'm doing here is basically to understand what arguments were passed to `<DocsLink>` and then invoking the LinkTo component with only those arguments.

I think this is still worth it since:
- we're only using public apis, so we should be fine
- we're full glimmer components
- we're not relying on `@ember/legacy-built-in-components`

So, this seems like a more future-proof approach to me. 
I think `<LinkTo>` should validate the arguments **values** and not just its presence. Basically treating an undefined argument as if it wasn't passed in. That would make the implementation a lot cleaner. But that's a another discussion and maybe a good issue for the ember.js repo.

addon/components/docs-link/index.hbs
```diff
@@ -0,0 +1,27 @@
...
+}}
+  <DocsLinkTo
+    class="docs-md__a"
+    @query={{or @query (hash)}}
```
miguelcobain, line 19: Passing in an empty hash seems to trick `<LinkTo>` to behave as if nothing was passed in.
Passing in undefined triggers an error.

rwwagner90 (approved):
This looks good, thanks @miguelcobain! Nice to see we are removing @ember/legacy-built-in-components

PR #1133, opened 2022-02-12
update markdown rendering approach

miguelcobain:
What this PR does:
- updates `marked` and `highlight.js` to the latest versions
- changes the approach to markdown parsing

Previously we were parsing the markdown with the vanilla tokenizer from marked. This meant that we had to make some massaging of the tokens to successfully remove extra paragraphs when inside a curly or angle bracket syntax (see `compactParagraphs` function).

This new proposal makes use of marked's extensions, trying to accomplish a similar same thing.

One advantage is that the content **inside** an html or hbs block now is *preserved*, including whitespace.
As an example, previously compiling this markdown:

```markdown
{{#foo-bar}}

    hello

{{/foo-bar}}
```

resulted in the following hbs:

```hbs
<div class="docs-md"><p>{{#foo-bar}} hello {{/foo-bar}}</p></div>
```

But now it results in the the following:
```hbs
<div class="docs-md">
<p>{{#foo-bar}}

    hello

{{/foo-bar}}</p>
</div>
```

Notice how everything inside `{{#foo-bar}}` and `{{/foo-bar}}` is preserved. Same thing happens for html/angle bracket components.
This explains why I had to change tests.

Test coverage was good, but I went manually to every single page in the docs app and everything looks good.

Since @dfreeman seems to have worked on this code in the past, I'm going to mention him. Would be great to get another pair of eyes, at least in the test changes. :)


addon/components/docs-code-highlight/index.js
```diff
@@ -5,7 +5,6 @@ import hljs from 'highlight.js/lib/core';
 import javascript from 'highlight.js/lib/languages/javascript';
 import css from 'highlight.js/lib/languages/css';
 import handlebars from 'highlight.js/lib/languages/handlebars';
-import htmlbars from 'highlight.js/lib/languages/htmlbars';
```
miguelcobain, line 8: `htmlbars` is no longer a thing in highlight.js, being replaced with just `handlebars`.

addon/components/docs-code-highlight/index.js
```diff
@@ -28,6 +27,6 @@ hljs.registerLanguage('ts', typescript);
 
 export default class DocsCodeHighlight extends Component {
   setupElement(element) {
-    hljs.highlightBlock(element);
```
miguelcobain, line 31: `highlightBlock` method was deprecated, being replaced with `highlightElement`.

SergeAstapov (approved):

PR #1137, opened 2022-02-14
fix(config): use runtime config instead of embroider generated one

jkeen:
Without this fix, addon docs deploys won't load correctly when using `ember-auto-import >= 2`. For me instead of loading versions.json from `https://ember-stereo.com/docs/ember-stereo.json` upon page load, it instead tried to load from `https://ember-stereo.com/ADDON_DOCS_ROOT_URL/docs/ember-stereo.json`

### DIFF
--- a/addon/adapters/-addon-docs.js
+++ b/addon/adapters/-addon-docs.js
@@ -1,10 +1,15 @@
+import { getOwner } from '@ember/application';
 import Adapter from '@ember-data/adapter';
-import config from 'ember-get-config';
 import fetch from 'fetch';
 
 export default class AddonDocsAdapter extends Adapter {
   defaultSerializer = '-addon-docs';
-  namespace = `${config.rootURL.replace(/\/$/, '')}/docs`;
+
+  get namespace() {
+    const rootURL =
+      getOwner(this).resolveRegistration('config:environment').rootURL;
+    return `${rootURL.replace(/\/$/, '')}/docs`;
+  }
 
   shouldBackgroundReloadAll() {
     return false;


rwwagner90 (approved):
@jkeen this looks good, thanks! I wonder if we need to do the same thing elsewhere to replace more usages of ember-get-config?

anehx:
Thanks for opening the PR @jkeen ! I forgot it :wink:

jkeen:
Thanks for figuring it out, @anehx! 

And yeah, maybe re: replacing ember-get-config everywhere, @rwwagner90? Not certain though, as everything seems to be functioning properly now (on ember-stereo.com anyway) as far as I can tell.

rwwagner90:
@jkeen we definitely don't have embroider support. I have a PR open to test embroider, and it seems to fail https://github.com/ember-learn/ember-cli-addon-docs/pull/1136

Do you or @anehx have any ideas on how we fix embroider compatibility? Are you using embroider for your builds?

SergeAstapov:
@jkeen was you able to understand the root cause of this problem? `ember-get-config` should be be working just fine in other projects. I wonder if an issue could be caused by mix of `ember-get-config` and `@emroider/macros` in the deps tree.

@rwwagner90 I'm pretty sure this change should not have any effect on embroider support, as embroider build fails on another issue - it does not pull add-on files into build (still haven't figured out the root cause for this)

rwwagner90:
@SergeAstapov hmm, I was just speculating since the title of this PR was `use runtime config instead of embroider generated one` thought maybe it helped us with embroider.

jkeen:
Sorry fellas, I just found the commit on @anehx's repo after a deep search and just used his commit message for the PR. My understanding of underlying causes of the issue are pretty high level in that it was broken before and this fixed it. 🤷🏻‍♂️ 

(I'm also having one hell of a time making this jump to ember-auto-import 2 + embroider in addons and apps I'm working on so I figured the least I could do was connect couple of dots and make a PR to hopefully save someone else some hassle if they run into this same cryptic issue.)

anehx:
@rwwagner90 @SergeAstapov I don't think replacing `ember-get-config` would resolve any of the embroider issues we have. However, I'd still do it since the recommended way to use the config is via the container ([see comment in readme of `ember-get-config`](https://github.com/mansona/ember-get-config)). Now, to answer the question about the root cause:

`ember-get-config` uses `@embroider/macros` function `getOwnConfig` which will create a copy of the config in your source code instead of using the meta tag or the config file if configured so. `ember-cli-addon-docs` forces the host apps to set `rootURL` to `"/ADDON_DOCS_ROOT_URL/"` and then replaces that placeholder later on. However, this only affects the runtime configuration but not the copy of the config which embroider created. When we used `rootURL` (with `ember-get-config`) in the adapter it used `/ADDON_DOCS_ROOT_URL/` as namespace for the request.

rwwagner90:
@anehx I think we have some usages outside the container, that require ember-get-config or possibly some other magic, but if I am wrong about that, I would love some help doing everything the right way 😃

anehx:
@rwwagner90 there you go: #1143

SergeAstapov:
> @rwwagner90 @SergeAstapov I don't think replacing `ember-get-config` would resolve any of the embroider issues we have. However, I'd still do it since the recommended way to use the config is via the container ([see comment in readme of `ember-get-config`](https://github.com/mansona/ember-get-config)). Now, to answer the question about the root cause:
> 
> `ember-get-config` uses `@embroider/macros` function `getOwnConfig` which will create a copy of the config in your source code instead of using the meta tag or the config file if configured so. `ember-cli-addon-docs` forces the host apps to set `rootURL` to `"/ADDON_DOCS_ROOT_URL/"` and then replaces that placeholder later on. However, this only affects the runtime configuration but not the copy of the config which embroider created. When we used `rootURL` in the adapter it used `/ADDON_DOCS_ROOT_URL/` as namespace for the request because it consumed the config copy generated by `ember-get-config`.

Thanks for explanation! Now this makes sense! Basically `ember-get-config` v1+ does not honor anything addons add into config via `config()` hook e.g. [everything here](https://github.com/ember-learn/ember-cli-addon-docs/blob/master/index.js#L54-L109) unless the addon runs *before* `ember-get-config`

PR #1164, opened 2022-03-11
Support v2 addon format

SergeAstapov:
The `src` folder is a the new convention for the location of the addon source code.

This can be seen in [Porting an Addon to V2](https://github.com/embroider-build/embroider/blob/main/PORTING-ADDONS-TO-V2.md) guide as well as part of [new addon blueprint](https://github.com/embroider-build/addon-blueprint/tree/main/files/packages/__name__).

This was tested in https://github.com/ember-animation/ember-animated/pull/397

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,11 +72,11 @@ module.exports = {
     let addonPathInRepo = this._documentingAddonAt()
       ? path.relative(
           this._getRepoRoot(),
-          path.join(this._documentingAddonAt(), 'addon')
+          path.join(this._documentingAddonAt(), this._addonSrcFolder())
         )
       : path.relative(
           this._getRepoRoot(),
-          path.join(this.project.root, 'addon')
+          path.join(this.project.root, this._addonSrcFolder())
         );
 
     let config = {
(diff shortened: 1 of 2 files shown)


rwwagner90:
@SergeAstapov is this all that is required? Does moving an addon to V2 fix our issues with embroider compatibility or does this just work for ember-animated because it has the docs as a separate app from the tests?

SergeAstapov:
> is this all that is required?

yes, as far as I can see working on `ember-animated` - this is all what is needed from `ember-cli-addon-docs` package.

> Does moving an addon to V2 fix our issues with embroider compatibility?

unfortunately no. There are some issues in the `ember-cli-addon-docs` or it's dependencies. still haven't figured out what exactly.

> does this just work for ember-animated because it has the docs as a separate app from the tests?

exactly! the docs app in `ember-animated` is not being tested against embroider

ef4:
I will point out that for v2 addons this is only a convention and not a hard requirement. So this PR makes sense as a good default, but it would also be good to let addon authors set this path explicitly.

rwwagner90:
@SergeAstapov @ef4 do we want to allow users to tweak the path as part of this PR or should we follow up later? I am fine either way.

SergeAstapov:
@rwwagner90 I'd like to introduce that ability in this PR. It should be relatively simple and no need to have multiple PRs to introduce such feature.

@ef4 thank you for the idea making this configurable.

Converted PR to draft for now.

SergeAstapov:
@rwwagner90 updated PR: `src` is now default folder name for the v2 addon source code which could be overridden via `addonSrcFolder` like so:

```js
let app = new EmberApp(defaults, {
  'ember-cli-addon-docs': {
    documentingAddonAt: '../addon',
    addonSrcFolder: 'source',
  }
});
```

rwwagner90 (approved):

PR #1276, opened 2022-07-02
avoid using `this.set` (fixes #1275)

### LINKED ISSUE #1275: DocsKeyboardShortcutsComponent: pressing 'g' key throws TypeError
When I run the ember-intl docs app locally (uses v5.0.0 of ember-cli-addon-docs) and then press the 'g' key, I see an error logged to the console:

```
Uncaught TypeError: this.set is not a function
    at ClassAsKeyResponder.goto (index.js:247:1)
    at eval (handle-key-event.js:16:1535)
    at Array.forEach (<anonymous>)
    at triggerResponderListener (handle-key-event.js:16:1354)
    at handleKeyEventWithPropagation (handle-key-event.js:16:897)
    at eval (keyboard.js:29:1907)
    at Backburner._run (backburner.js:1052:1)
    at Backburner.run (backburner.js:773:1)
    at run (index.js:83:1)
    at KeyboardService._respond (keyboard.js:29:1769)
```

This makes sense, since `DocsKeyboardShortcutsComponent` is a glimmer component which won't have `this.set` available. I think that we can probably just switch this over to a regular assignment to avoid the problem.

geneukum:
Right now, the DocsKeyboardShortcutsComponent attempts to use `this.set`
to set a property. DocsKeyboardShortcutsComponent is a Glimmer
Component, so this API is not available.  To avoid the problem, let's
use a regular assignment statement.

### DIFF
--- a/addon/components/docs-keyboard-shortcuts/index.js
+++ b/addon/components/docs-keyboard-shortcuts/index.js
@@ -22,9 +22,9 @@ export default class DocsKeyboardShortcutsComponent extends Component {
   @onKey('KeyG', { event: 'keyup' })
   goto() {
     if (!formElementHasFocus()) {
-      this.set('isGoingTo', true);
+      this.isGoingTo = true;
       later(() => {
-        this.set('isGoingTo', false);
+        this.isGoingTo = false;
       }, 500);
     }
   }


addon/components/docs-keyboard-shortcuts/index.js
```diff
@@ -22,9 +22,9 @@ export default class DocsKeyboardShortcutsComponent extends Component {
   @onKey('KeyG', { event: 'keyup' })
   goto() {
     if (!formElementHasFocus()) {
-      this.set('isGoingTo', true);
+      this.isGoingTo = true;
```
rwwagner90, line 25: Does `isGoingTo` need to be a tracked prop?

geneukum: I don't think so. 

My understanding is that `isGoingTo` is basically just an internal flag that we're using to chain together multiple keyup events for a keyboard shortcut rather than something that affects how the component is rendered.

rwwagner90 (approved):

Issue #1415, opened 2023-04-10
Not GDPR compliant out of the box


mixonic:
Addon docs uses Google fonts via https://github.com/ember-learn/ember-cli-addon-docs/blob/b330f9af6e159655aba360bee48f646dd382ad7a/index.js#L227

As Google may use their fonts service to analyze user behavior (at the least gathering IP addresses), and there is no clear way for a user to opt-out, addon docs websites may not be GDPR compliant out of the box.

This could be resolved by having the site serve the font asset, not using a custom font at all, or by finding a GDPR compliant font hosting service.

At the least, a documented workaround should be available.

mixonic:
@SergeAstapov @RobbieTheWagner @miguelcobain I'm happy to solve this any of a few ways if you can give me a rough approval for what you think reasonable. Options available seem like:

1. Add a build-time configuration that allows users to disable the font link. Add this to the docs. (requires a minor release)
2. Change the default fonts to be appropriate system serif/sans fonts, and describe in the docs how to set a custom font. (would be a major release)
3. Download and check the font into the assets of the theme so they come from the same domain instead of Google. (requires a minor release)

RobbieTheWagner:
@mixonic addon docs is mostly unmaintained now, since it won't work with v2 addons, embroider, or fastboot. If anyone wants to put in some time to help with those things, I would love to keep it going, but as of right now, there isn't really a path forward.

mixonic:
@RobbieTheWagner I'll keep that in mind, but I'd still love to address this issue upstream instead of forking for it. I'll take it your vote as "do the least effort thing unless you are fixing one of these other three issues as well".

RobbieTheWagner:
Yeah, that's an accurate assessment. I'm fine with any fix.

PR #1421, opened 2023-04-22
Swap out 3rd party font for system fonts

mixonic:
Ember CLI Addon Docs should use system fonts by default. System fonts load quickly for users, reduce the site's coupling to 3rd party assets, and make sites privacy-friendly by default.

The Google font previously relied upon would result in Google having the ability to add cookies for that traffic, possibly placing a site out of compliance with EU and other regulations regarding user tracking without consent processes in place.

Here I've replaced (via https://github.com/Addepar/ember-table/pull/983 for the purposes of illustration):

* https://fonts.google.com/specimen/Crimson+Text
* Try it in your browser at https://opensource.addepar.com/ember-table/versions/v5.0.3/docs

Screenshot of previous font on macOS
Screenshot of previous font on Ubuntu

with

* https://github.com/system-fonts/modern-font-stacks#old-style
* Try it in your browser at https://opensource.addepar.com/ember-table/versions/master/docs

Screenshot of new font on macOS
Screenshot of new font on Ubuntu

Note that the strategy of font stacks is to pick a class of font, so this will render differently on Linux vs. macOS, etc.

I'm not sure what kind of release this warrants, but I'm going to suggest **minor**. It is not a breaking change, though it is a visual change.

Fixes https://github.com/ember-learn/ember-cli-addon-docs/issues/1415

### DIFF
--- a/addon/styles/tailwind.config.js
+++ b/addon/styles/tailwind.config.js
@@ -39,7 +39,7 @@ let textSizesMap = {
   'large-7': '42px',
   'jumbo-1': '60px',
   'jumbo-2': '74px',
-  'jumbo-3': '86px'
+  'jumbo-3': '86px',
 };
 
 // Convert pixel sizes to REMs
@@ -88,7 +88,7 @@ module.exports = {
       'code-base': '#282c34',
 
       // See the note in tailwind/compnents/docs-brand-colors.css
-      'brand-var': 'var(--brand-primary, #E04E39)'
+      'brand-var': 'var(--brand-primary, #E04E39)',
     },
 
     /*
@@ -114,7 +114,7 @@ module.exports = {
       sm: '576px',
       md: '768px',
       lg: '992px',
-      xl: '1200px'
+      xl: '1200px',
     },
 
     /*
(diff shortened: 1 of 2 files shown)


mixonic:
I'm taking the initiative to land this based on the consensus that any solution was acceptable expressed at https://github.com/ember-learn/ember-cli-addon-docs/issues/1415.

Ping me if there is anything to follow up on! I'll look into triggering a release in the next day or two.

addon/styles/tailwind.config.js
```diff
@@ -137,7 +137,16 @@ module.exports = {
...
-      title: ['"Crimson Text"', 'serif'],
+      /*
+       * https://github.com/system-fonts/modern-font-stacks#old-style
+       */
+      title: [
+        '"Iowan Old Style"',
+        '"Palentino Linotype"',
+        '"URW Palladio L"',
+        '"P052"',
+        'serif',
```
c69-addepar, line 148: With `serif` fallback, it should work pretty much anywhere, but still - maybe its worth adding a windows font to the fallback list ?

c69-addepar (approved):
LGTM but small comment

addon/styles/tailwind.config.js
mixonic: Per modern font stacks, `Palatino Linotype` is available on Windows XP and up.

mixonic:
@RobbieTheWagner I haven't been able to make a release cut yet, so please feel free to provide feedback and I will address it!

I'm also blocked on making a release until I'm granted permissions on the npm package, if you can unblock that. (@dfreeman was going to help me out)

PR #1454, opened 2023-06-22
embroider ready

patricklx:
this is a way to support embroider and classic build.

as postprocessTree does not exists for embroider, alternative ways where needed.

The drawback.of this could be that if some other type of files would be converted to .md files as those would not be picked up.
And other files that could be converted to hbs files


SergeAstapov:
@patricklx looks like CI is unhappy with proposed changes :/

patricklx:
@SergeAstapov try again

patricklx:
Ah, lock file issue and linter... Will fix it later

patricklx:
fixed lock file issue

patricklx:
i had to move babel.config.js inside eslint.rc file as it was breaking embroider build

patricklx:
it should now all work

ef4 (approved):
Replacing postprocessTree('all') with treeForPublic and treeForAddon is good.

SergeAstapov (approved):

RobbieTheWagner (approved):
LGTM! I thought we had no way to support embroider

PR #1485, opened 2023-09-07
Add ember-tether to dependencies

### LINKED ISSUE #1484: Version dropdown doesn't work when using pnpm
I switched to pnpm in my addon. When I run the docs site, clicking on the versions dropdown in the main header gives me the following error:

Screenshot 2023-09-07 at 12 33 37 PM

I was able to resolve this by installing `ember-tether` in my test-app. But I think this should be fixed in `ember-cli-addon-docs` by declaring `ember-tether` as a dependency.

barryofguilder:
This adds `ember-tether` to the package "depencencies" and will fix #1484.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "ember-router-scroll": "^4.1.2",
     "ember-set-helper": "^2.0.1",
     "ember-svg-jar": "^2.4.4",
+    "ember-tether": "^3.0.0",
     "ember-truth-helpers": "^3.0.0",
     "esm": "^3.2.25",
     "execa": "5.1.1",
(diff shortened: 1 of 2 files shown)


barryofguilder:
Hmm, with pnpm, I still get the following error unless the consuming test-app specifies `ember-tether` in it's dependencies. I'm not great with dependency resolution, so not sure what needs to be changed.

Screenshot 2023-09-07 at 12 33 37 PM

RobbieTheWagner:
@barryofguilder thanks for the PR! I think what you have here is correct because `ember-tether` is a peerDep of `ember-modal-dialog`. Let's try merging and releasing this and see if the released version works for you.

PR #1517, opened 2023-12-20
fix compatibility with ember-data 5

### LINKED ISSUE #1507: Error from ember data with the module model
Hello, while upgrading an addon, i got this error from ember-data : 

```
Uncaught Error: The schema for the relationship 'component.components' is not configured to satisfy 'module' and thus cannot utilize the 'module.components' relationship to connect with 'class.implicit-module:components1700295419147'

If using this relationship in a polymorphic manner is desired, the relationships schema definition for 'component' should include:

{
  components: {
    name: 'components',
    type: 'class',
    kind: 'hasMany',
    options: {
      as: 'undefined', <---- should be 'module'
      async: false,
      polymorphic: false,
      inverse: 'implicit-module:components1700295419147'
    }
  }
}

 and the relationships schema definition for 'class' should include:

{
  implicit-module:components1700295419147: {
    name: 'implicit-module:components1700295419147',
    type: 'module',
    kind: 'implicit',
    options: {
      as: 'undefined', <---- should be 'class'
      async: false,
      polymorphic: false, <---- should be true
      inverse: 'components'
    }
  }
}

    at assertInheritedSchema (-private.js:423:1)
    at assertConfiguration (-private.js:656:1)
    at isLHS (-private.js:673:1)
    at Graph.get (-private.js:1685:1)
    at addToInverse (-private.js:1108:1)
    at replaceRelatedRecordsRemote (-private.js:1067:1)
    at replaceRelatedRecords (-private.js:946:1)
    at Graph.update (-private.js:1894:1)
    at updateRelationshipOperation (-private.js:1535:1)
    at Graph.update (-private.js:1863:1)
```

Doing a patch of the module model made it work: 

```
diff --git a/addon/models/module.js b/addon/models/module.js
index 4887ba3bd30829653cd6944f1097ed65db41a9da..8c10bc597ef794f99e52549d0bef31eae12b08a1 100644
--- a/addon/models/module.js
+++ b/addon/models/module.js
@@ -13,7 +13,7 @@ export default class Module extends Model {
   @hasMany('class', { async: false, inverse: null })
   classes;
 
-  @hasMany('class', { async: false, inverse: null })
+  @hasMany('component', { async: false, inverse: null })
   components;
 
   /*
```

I don't really know if the `components` property of the `module` model should be of type 'component' or 'class'. 

If you want to reproduce the error, here is the pull-request: https://github.com/shipshapecode/ember-flatpickr/pull/1911. You just need to remove 
```
"pnpm": {
    "patchedDependencies": {
      "ember-cli-addon-docs@6.0.2": "patches/ember-cli-addon-docs@6.0.2.patch"
    }
  }
```
from the package.json.

I'd be happy to make a PR if needed.

AmauryD:
Fixes #1507

### DIFF
--- a/addon/models/module.js
+++ b/addon/models/module.js
@@ -10,10 +10,10 @@ export default class Module extends Model {
   @attr
   functions;
 
-  @hasMany('class', { async: false, inverse: null })
+  @hasMany('class', { async: false, inverse: null, polymorphic: true })
   classes;
 
-  @hasMany('class', { async: false, inverse: null })
+  @hasMany('class', { async: false, inverse: null, polymorphic: true })
   components;
 
   /*
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -98,6 +98,7 @@ jobs:
         try-scenario:
           - 'ember-lts-4.4'
           - 'ember-lts-4.8'
+          - 'ember-lts-5.4'
           - 'ember-release'
           - 'ember-beta'
           - 'ember-canary'
--- a/addon/models/class.js
+++ b/addon/models/class.js
@@ -3,7 +3,7 @@ import { filterBy, or, union } from '@ember/object/computed';
 import { memberUnion, hasMemberType } from '../utils/computed';
 
 export default class Class extends Model {
-  @belongsTo('class', { async: false, inverse: null })
+  @belongsTo('class', { async: false, inverse: null, polymorphic: false })
   parentClass;
 
   isClass = true;
--- a/addon/models/project.js
+++ b/addon/models/project.js
@@ -13,6 +13,6 @@ export default class Project extends Model {
   @attr
   navigationIndex;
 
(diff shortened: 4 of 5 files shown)


addon/models/module.js
```diff
@@ -10,10 +10,10 @@ export default class Module extends Model {
   @attr
   functions;
 
-  @hasMany('class', { async: false, inverse: null })
+  @hasMany('class', { async: false, inverse: null, polymorphic: true })
   classes;
 
-  @hasMany('class', { async: false, inverse: null })
+  @hasMany('class', { async: false, inverse: null, polymorphic: true })
```
AmauryD, line 16: If someone knowledgeable about this library comes across this, is it normal for 'components' to be a 'class' model rather than a 'component' model ?

SergeAstapov (approved):
Thank you!

RobbieTheWagner (approved):
Seems reasonable to me, thanks for the fix!

PR #1582, opened 2024-07-30
Fix @ember/string dependency

anehx:
`@ember/string` must be a dependency as `ember-cli-addon-docs` imports it in their codebase. Currently this still works, but with v4 of `@ember/string` (v2 addon) this will break as it fully relies on `ember-auto-import`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,6 +155,7 @@
     "webpack": "^5.89.0"
   },
   "peerDependencies": {
+    "@ember/string": "^3.1.1 || ^4.0.0",
     "ember-data": ">= 3.0.0",
     "ember-fetch": "^8.1.1",
     "ember-source": ">= 4.0.0"


RobbieTheWagner:
@anehx should it be a peerDep?

anehx:
@RobbieTheWagner I don't think so. This addon is mostly used in other addons which do not include `@ember/string` in the `devDependency` in their blueprint so the users of this addon would have to install this manually..

RobbieTheWagner:
I think we should probably make it a peerDep and allow v3 and v4 like this https://github.com/minutebase/ember-can/blob/master/ember-can/package.json#L129

anehx:
@RobbieTheWagner Fair enough. I changed it in this PR.

package.json
```diff
@@ -105,7 +105,6 @@
     "@babel/plugin-proposal-decorators": "^7.23.6",
     "@babel/preset-env": "^7.23.6",
     "@ember/optional-features": "^2.0.0",
-    "@ember/string": "^3.1.1",
```
RobbieTheWagner, line 108: This will still need to be a devDep too right, so we're installing it ourselves when running the test app etc?

anehx: Absolutely, my bad. Fixed!

RobbieTheWagner (approved):

RobbieTheWagner:
CI is failing for other reasons, so I will go ahead and merge this. Thanks for the PR!

ijlee2:
Hi, @RobbieTheWagner. Could you help release a patch version of `ember-cli-addon-docs`?

The documentation site for `ember-intl` needs the fix in this PR to be able to update `@ember/string` to `v4`. A new version would also help me remove a custom patch that I had upstreamed in #1556.

RobbieTheWagner:
@ijlee2 I would love to help release a new version, but CI is failing currently. If you have time to diagnose and fix those failures, we could definitely get a new version out.

ijlee2:
@RobbieTheWagner Ah, I didn't see the CI failing in the main branch.

Since the error message refers to `@babel/plugin-transform-class-static-block`, I'd try updating Babel-related dependencies to the latest version. This may require additional maintenance work, though. Maybe we can look into it after EmberFest.

PR #1670, opened 2025-04-17
fix: handlebars parser@2.2.0

### LINKED ISSUE #1669: Problem with @handlebars/parser @ 2.2.0
Given app created like:

```bash
ember addon sandboxaddon
cd sandboxaddon/
pnpm install
pnpm exec ember install ember-cli-addon-docs
pnpm exec ember install ember-cli-addon-docs-yuidoc
pnpm start
```

On the last step I will get following error:

```
> sandboxaddon@0.0.0 start /Users/michal/sandbox/sandboxaddon
> ember serve

require() of ES Module /Users/michal/sandbox/sandboxaddon/node_modules/.pnpm/@handlebars+parser@2.2.0/node_modules/@handlebars/parser/dist/cjs/index.js not supported.
Instead change the require of index.js in null to a dynamic import() which is available in all CommonJS modules.


Stack Trace and Error Report: /var/folders/8q/yrjx_8s115q43r8w5qd90_m00000gn/T/error.dump.374c2f832c85fd615c1e5d4c2fc23452.log
 ELIFECYCLE  Command failed with exit code 1.
```

This is somehow related to `v2.2.0` of `@handlebars/parser` because when I add following to `package.json`:

```json
  "pnpm": {
    "overrides": {
     "@handlebars/parser": "2.1.0"
    }
  }
```

And then re-start the process:

```bash
pnpm install
pnpm start
```

Then this error goes away.

MichalBryxi:
- Fixes #1669


.github/workflows/addon-docs.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           cache: pnpm
       - name: Install Dependencies
         if: env.SHORT_CIRCUIT == 'false'
-        run: pnpm install --no-lockfile
+        run: pnpm install
```
MichalBryxi, line 44: This is always a good idea.

MichalBryxi:
Despite GitHub insisting that there is some diff, I see no diff from this branch against `main` so I would just close this.

PR #1672, opened 2025-04-23
fix: Remove ember-cli-string-helpers

### LINKED ISSUE #1666: Remove `ember-cli-string-helpers` addon
When I upgrade my addon to use `@ember/string` v4.0, my addon build will throw an error that says:

"Module not found: Error: ember-cli-string-helpers is trying to import the app's @ember/string package, but it seems to be missing"

I noticed that the `ember-cli-string-helpers` addon was archived in October 2024. I think it would be a good idea to remove this package and replace it with something that is maintained. For now, the way to fix this issue is by locking the version of `@ember/string` to 3.1.1.

MichalBryxi:
## Core
- This pulls @ember/string 3.1.1 into dependent project and that causes a lot of issues
- Since we have `capitalize` in @ember/string, we can just call that in our internal helper
- ~~Also bumped internal version of @ember/string to v4.x~~ -> I could not make it work, next step

## Fixed tests
- Also pins portal target for ember-tether to correct scope for testing so that assertions can work correctly

Fixes #1666

### DIFF
--- a/addon/helpers/capitalize.js
+++ b/addon/helpers/capitalize.js
@@ -1,0 +1,6 @@
+import { helper } from '@ember/component/helper';
+import { capitalize } from '@ember/string';
+
+export default helper(function capitalizeHelper(positional) {
+  return capitalize(positional[0]);
+});
--- a/app/helpers/capitalize.js
+++ b/app/helpers/capitalize.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-cli-addon-docs/helpers/capitalize';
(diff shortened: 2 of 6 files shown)


MichalBryxi:
Anywho: Ready for review @SergeAstapov

addon/components/docs-viewer/x-nav/index.hbs
```diff
@@ -1,32 +1,26 @@
-{{#if this.media.isMobile}}
```
SergeAstapov, line 1: @MichalBryxi did this leak somehow as pr title only says about `ember-cli-string-helpers`?

MichalBryxi:
Ok so now this PR works. It only does:

1) Remove `ember-cli-string-helpers`
2) Fix `ember-tether` tests not finding correct container (not sure how it worked before)

Ready for merge @SergeAstapov

SergeAstapov (approved):

MichalBryxi:
Yay \o/

PR #1679, opened 2025-05-02
fix: Do not override ember-modal-dialog components

MichalBryxi:


addon/components/docs-header/search-results/index.hbs
```diff
@@ -1,6 +1,10 @@
-<div {{did-insert (perform this.search)}} {{did-update (perform this.search) @query}} ...attributes>
+<div
```
MichalBryxi, line 1: Not exactly sure why my IDE wants to reformat everything in this project. Unless I'm missing some settings and actually my IDE is misconfigured, then might as well. Happy to revert if you think it makes too much noise.

SergeAstapov: @MichalBryxi seems like you have prettier

SergeAstapov (approved):

SergeAstapov:
Thank you @MichalBryxi !

PR #1668, opened 2025-04-13
docs: How do I "add ecad"?

MichalBryxi:
- Adding section does not have any isntructions.
- Should I just `ember install`? Or are there any other steps?

PR #1681, opened 2025-05-02
fix: Broken link to quickstart

MichalBryxi:
- Fixes: #1668


SergeAstapov:
thank you @MichalBryxi for quick fix!

PR #1709, opened 2026-08-02
migrate from inject to service named import in @ember/service

jelhan:
The `service` named export has been added in [v4.1](https://blog.emberjs.com/ember-4-1-released/) as an alternative to `inject`. The `inject` named export was [deprecated](https://deprecations.emberjs.com/id/importing-inject-from-ember-service/) in the v6 cycle and dropped in 7.0.

This should not be a breaking change as ember-cli-addon-docs already requires Ember v4.8+.

This is required for Ember v7 compatibility. But I have not tested if it is the only needed change.


MinThaMie (approved):
Thank you so much for this work!

